PR #1, opened 2020-07-31
Build and test without Redis

andrewdavidmackenzie:
Avoid a hard dependency on redis to be able to build and pass tests


davidor:
I've rebased this on top of master. I've also clean things up a bit by using the cfg-if crate.
All the tests pass without Redis except the doc ones. Not sure how to pass them conditionally without modifying the examples in the docs.

PR #5, opened 2020-11-13
Switch to Envoy RLS protocol v3

davidor:
v2 is deprecated. Given that this is a new project, we don't need to support v2 and v3.

Let's wait until [marin3r](https://github.com/3scale/marin3r) is ready. @roivaz , @slopezz let me know when we should merge.


limitador-server/src/envoy_rls/envoy_types.rs
```diff
@@ -39,9 +27,6 @@ pub mod envoy {
 
     pub mod service {
         pub mod ratelimit {
-            pub mod v2 {
-                include!("protobufs/envoy.service.ratelimit.v2.rs");
-            }
             pub mod v3 {
                 include!("protobufs/envoy.service.ratelimit.v3.rs");
```
unleashed, line 31: Not directly related, but _since we are here_ you can look at using `#[path = "..."]` rather than `include!` for slightly less lines and compatibility with rust-analyzer. [Example](https://github.com/3scale/gateway-ng-controller/blob/11757dbd6f689adc4112fb867e1a0290e70964fd/src/protobuf.rs).

unleashed (approved):
LGTM. You might want to test this with an instance of envoy launched with `--bootstrap-version 3` to ensure most (all?) things default to v3.

limitador-server/src/envoy_rls/envoy_types.rs
davidor: It's not slightly less lines in this case because there are multiple nested mods. The first part would be like this:
```rust
#[path = "protobufs"]
pub mod envoy {
    #[path = "."]
    pub mod config {
        #[path = "."]
        pub mod core {
            #[path = "envoy.config.core.v3.rs"]
            pub mod v3;
        }
    }
```
Intellij seems to parse the first option just fine, but I can change it if it causes issues in your setup.

unleashed: Actually in RA `include!` works for autocompletion if it is using only a relative path, but at a minimum go-to-definition fails. With #[path] it works just fine. Does the latter work well in intellij-rust?

davidor: Added in new commit :+1:

davidor:
[marin3r](https://github.com/3scale/marin3r)  switched to v3 so this can be merged.

davidor:
@unleashed I forgot to delete v2 from `build.rs`. I've added a new commit with that.

unleashed (approved):

PR #7, opened 2020-12-11
Update k8s templates to Limitador-server v0.3.0

davidor:
- Updates limitador-server to 0.3.0
- Updates the Kuard envoy config to use Envoy v3. This is needed because limitador-server 0.3.0 is only compatible with v3 of the rls protocol. While at it, I updated the other filters too (connection manager, and access logs).
- Changed the image pull policy to "IfNotPresent". We needed "Always" before we didn't have stable releases.


slopezz (approved):
I have tested it on AWS, and it works as expected!

PR #8, opened 2020-12-11
Github workflow: Run in KinD

davidor:
Adds a new step to the Github Worklow that runs Limitador in KinD. This is useful to check that we didn't break anything in the kubernetes templates.


slopezz (approved):
As a first step on having better e2e test is ok!

PR #10, opened 2021-01-12
Dockerfile: do not set LIMITS_FILE

davidor:
Limitador shouldn't load any limits by default. This was done for testing purposes and I forgot to change it.


unleashed (approved):
bors r+

davidor:
Rebased on top of master to pick fixes introduced in #11

PR #11, opened 2021-01-12
Fix warnings included in the new version of Clippy

davidor:


limitador-server/src/http_api/mod.rs
```diff
@@ -1,2 +1,7 @@
+// Deriving ApiV2Schema raises this Clippy warning. We can delete this once we
+// upgrade to a version of paperclip that contains the fix
+// https://github.com/wafflespeanut/paperclip/pull/275
+#[allow(clippy::field_reassign_with_default)]
```
unleashed, line 4: I don't know how to allow this specifically for the `derive` inside. This looks good enough anyway.

unleashed (approved):

PR #12, opened 2021-01-20
Update dependencies, including Redis to use Tokio 1.0

unleashed:
This just brings updated versions of direct dependencies where available, including a couple small fixes to adapt to these releases.

Passes all tests locally.

There is an issue with the `redis` feature flag requiring `tokio` that is not expressed in the manifests, but that can be fixed later.


davidor (approved):

davidor:
@unleashed the tests pass but the application fails to start.
This is the error: `thread 'main' panicked at 'there is no reactor running, must be called from the context of Tokio runtime`

Issue #22, opened 2021-02-22
Add name attribute in Limit


davidor:
Adding a name to the limit and exposing it in the Prometheus metrics, would allow users to show information about a particular rate-limit identified by a "user-friendly" name in Grafana dashboards.

davidor:
@jmelis @jhernand I started working on this feature and I have a more specific proposal on how to implement it.

First, using a limit name as a label could be problematic because there could be many different values. Check the warning in the Prometheus docs: https://prometheus.io/docs/practices/naming/#labels However, I think we can have the feature disabled by default and enable it via env. I suspect in some use cases there will only be a few limits and this shouldn't be a problem. How many limits do you think you'll have a single instance of Limitador? Do you think it could be a problem in your case?

Apart from that, for a single request limitador might check multiple limits. It returns rate-limited when it finds the first limit that's over the defined value. I think that adding the name of the limit as a label to the "denied" counter is valuable so you know which limit is responsible for denying the request. Notice that this would only increase the counter for the first limit that Limitador finds. We can't really increase the counter for all, that would imply checking all the limits unnecessarily. Would that be OK for your use case?

Regarding authorized request, I don't think it's so interesting to have the "metric name" label for them. There could be multiple limits checked so we'd need to increase several counters and I think it could even be a bit confusing. I think that normally we're just interested in knowing which limits are denying requests, but I'd like to hear your opinion about this.

jhernand:
I think that the key word of that warning in the Prometheus documentation is "unbounded". In this case the set of limit names will be bounded, and actually rather small. In the OCM clusters management service, for example, we won't probably ever have more limit names than URL patterns, and we have only 59 o those (and only one limit at the moment). We are already using those URL patterns in other Prometheus metrics and works fine (when done properly).

In our use case it is perfectly OK to only increase the counter of the first limit that is over the defined value.

I think that having the label in the authorized calls metric is interesting because then we can graph both together so that the human user can easily compare the authorized number of requests with the rejected number of requests. I mean, it is not the same to say "1 request rejected per second" than to say "90% of requests rejected". However I understand the difficulty to implement it. I think we can live without it, specially if we have the definition of the limit as labels or as additional metrics. I mean something like this:

```
limited_calls{limitador_namespace="myns",name="mylimit",max_value="60",seconds="60"} 12312
```

If we have that then we can report much more useful information to the user in graphs and alerts.

davidor:
Implemented in #25 and #26

[...]

PR #24, opened 2021-03-03
Refactor prometheus metrics adapter

davidor:
This is in preparation for the changes needed for #22 It will make things easier.

### DIFF
--- a/limitador/src/prometheus_metrics.rs
+++ b/limitador/src/prometheus_metrics.rs
@@ -3,6 +3,26 @@ use prometheus::{Encoder, IntCounterVec, IntGauge, Opts, Registry, TextEncoder};
 
 const NAMESPACE_LABEL: &str = "limitador_namespace";
 
+struct Metric {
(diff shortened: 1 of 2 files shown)


limitador/src/prometheus_metrics.rs
```diff
@@ -3,6 +3,26 @@ use prometheus::{Encoder, IntCounterVec, IntGauge, Opts, Registry, TextEncoder};
 
 const NAMESPACE_LABEL: &str = "limitador_namespace";
 
+struct Metric {
+    name: String,
+    description: String,
```
unleashed, line 8: these could very easily be &'a str since all contents are static strings

unleashed (approved):
LGTM

limitador/src/prometheus_metrics.rs
davidor: Right, but there are only 3 instances of this, so I don't think it's worth it.

PR #25, opened 2021-03-05
Add and optional name for limits

davidor:
This is part of #22.

This PR adds an optional name attribute for Limit. It also makes some changes in the Prometheus module to be able to include a limit name as a label when reporting a limited call.

Including the limit name as a label is optional. The reason is that it should be fine if the number of limits is low. However, with a large number of limits it could be problematic (check the caution note in the [Prometheus docs](https://prometheus.io/docs/practices/naming/#labels)).

This PR only addresses the part that concerns the lib. I'll do the server part the next PR.

### DIFF
--- a/limitador/src/prometheus_metrics.rs
+++ b/limitador/src/prometheus_metrics.rs
@@ -2,6 +2,7 @@ use crate::limit::Namespace;
 use prometheus::{Encoder, IntCounterVec, IntGauge, Opts, Registry, TextEncoder};
 
 const NAMESPACE_LABEL: &str = "limitador_namespace";
+const LIMIT_NAME_LABEL: &str = "limit_name";
 
 struct Metric {
     name: String,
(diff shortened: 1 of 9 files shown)


limitador/src/prometheus_metrics.rs
```diff
@@ -27,14 +28,52 @@ pub struct PrometheusMetrics {
...
+    use_limit_name_label: bool,
 }
 
 impl PrometheusMetrics {
     pub fn new() -> Self {
-        let labels = vec![NAMESPACE_LABEL];
+        Self::new_with_options(false)
+    }
+
+    pub fn new_with_counters_by_limit_name() -> Self {
```
davidor, line 39: I forgot to add a note here about why this is not enabled by default.

limitador/src/prometheus_metrics.rs
```diff
@@ -27,14 +28,52 @@ pub struct PrometheusMetrics {
...
+
+    pub fn incr_authorized_calls(&self, namespace: &Namespace) {
+        self.authorized_calls
+            .with_label_values(&[namespace.as_ref()])
+            .inc();
+    }
+
+    pub fn incr_limited_calls<'a, LN>(&self, namespace: &Namespace, limit_name: LN)
+    where
+        LN: Into<Option<&'a str>>,
```
unleashed, line 51: no `AsRef<str>` in the end?

unleashed (approved):
LGTM, check minor commments.

limitador/src/prometheus_metrics.rs
davidor: Added

limitador/src/storage/mod.rs
```diff
@@ -10,6 +10,11 @@ pub mod wasm;
 #[cfg(feature = "redis_storage")]
 pub mod redis;
 
+pub enum Authorization<'c> {
+    OK,
```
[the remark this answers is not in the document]
davidor: Right. Changed :+1:

limitador/src/prometheus_metrics.rs
davidor: In the end I decided not to do this because it makes the code more complicated (problems when inferring types, etc.).

PR #26, opened 2021-03-05
Add option in the server to enable using limit names in Prometheus labels

davidor:
Part of #22

This PR introduces a new env to enable using limit names in Prometheus labels. To do so, it uses the work done in the lib in the previous PR (#25)

### DIFF
--- a/limitador-server/docs/http_server_spec.json
+++ b/limitador-server/docs/http_server_spec.json
@@ -2,6 +2,7 @@
   "swagger": "2.0",
   "definitions": {
     "CheckAndReportInfo": {
+      "type": "object",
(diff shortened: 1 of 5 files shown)


limitador-server/docs/http_server_spec.json
```diff
@@ -2,6 +2,7 @@
   "swagger": "2.0",
   "definitions": {
     "CheckAndReportInfo": {
+      "type": "object",
```
davidor, line 5: @unleashed just in case you are not aware, this is generated automatically using https://github.com/wafflespeanut/paperclip

unleashed (approved):

PR #29, opened 2021-03-08
Release server 0.4.0

davidor:


davidor:
The "try in kind" check fails because it points to a tag in quay that will be created after this PR.

unleashed (approved):

PR #30, opened 2021-03-21
remove_generated_code

andrewdavidmackenzie:
Generate code for protobuf in "OUT_DIR" not under src/ and remove those files from version control.


limitador-server/src/envoy_rls/envoy_types.rs
```diff
@@ -4,8 +4,9 @@ pub mod envoy {
     pub mod config {
         #[path = "."]
         pub mod core {
-            #[path = "envoy.config.core.v3.rs"]
-            pub mod v3;
+            pub mod v3 {
+                include!(concat!(env!("OUT_DIR"), "/envoy.config.core.v3.rs"));
```
davidor, line 8: We switched from `include!` to `path` because @unleashed mentioned that `include!` does not work in some IDEs https://github.com/3scale-labs/limitador/pull/5#discussion_r534291152

davidor:
@andrewdavidmackenzie This PR breaks the Intellij integration for me. Intellij is not able to find the files generated from the protobufs and it shows in red all the structs like `Code`, `Entry`, etc. in `limitador-server/src/envoy_rls/server.rs`. is it the same for you or did you somehow fix it?

davidor:
`rustfmt` fails because there's a missing newline at the end of `limitador/src/storage/redis/redis_sync.rs`

andrewdavidmackenzie:
@davidor using #[path] instead of include!() works, but IntelliJ still cannot find the files, as it's looking in /src I assume.
Not sure if that is fixable, as they are generated in a build directory with a unique number, so you cannot even map it as a source file in IntelliJ.

alexsnaps:
I think this has been addressed as part of #104 - reopen if I'm wrong!

PR #32, opened 2021-04-13
Fix "get_limits()" integration test

davidor:
This PR fixes a test that was calling "get_counters()" instead of "get_limits()". As the title of the test indicates, the test should be about "get_limits". "get_counters" is tested a few lines below in the same file.


unleashed (approved):

davidor:
Rebased on top of master.

PR #35, opened 2021-04-28
Use #[path] instead of include!() to "include" generated files

andrewdavidmackenzie:
Add instructions on how to start redis for integration tests on mac os x
+ remove a trailing '/' from a link in readme that is warned on byy .md linter

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ port, that implements the Envoy Rate Limit protocol (v3).
 - [**Getting started**](#getting-started)
 - [**Limits storage**](#limits-storage)
 - [**Development**](#development)
-- [**Kubernetes**](limitador-server/kubernetes/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,6 +108,10 @@ Some tests need a redis deployed in `localhost:6379`. You can run it in Docker w
 ```bash
 docker run --rm --net=host -it redis
 ```
+on MacOS invoke Docker thus:
+```bash
+docker run --rm -p 6379:6379 -it redis
```
davidor, line 113: We can document this as the only option. It also works on linux and it's a better option. `--net=host` is more useful when you want to have connectivity between containers for example :+1:

davidor: The commit and PR messages don't match the changes you made

andrewdavidmackenzie:
I'll just do a new clean PR

PR #37, opened 2021-04-29
WIP - Add coverage with grcov

andrewdavidmackenzie:
Start by seeing that tests are OK if we switch to using rust nightly for Tests (only)


alexsnaps:
Mostly as an FYI, but since 1.60, this can now be done using llvm in stable: https://doc.rust-lang.org/rustc/instrument-coverage.html#test-coverage

PR #38, opened 2021-04-30
Infinispan limits storage (first version)

davidor:
This PR adds a limits storage implementation based on [Infinispan](https://infinispan.org/). It passes all the integration tests.

Infinispan does not support directly a few data types that we need, so I had to implement them with what Infinispan offers. Those data types are:
- Counters with TTLs.
- Sets.
- Locks (take a look at the limitations I documented in the code).

I added support for each of those in separate commits.

Notice that some of the operations are not very performant. Because of the lack of support in Infinispan, we need to serialize some things, make multiple requests for some operations, etc, but I think it should be fine for a first version.

Also, there are some things that could be improved if we added some supported operations in the infinispan rust client. For example a "reset" function for counters. I've left those improvements in TODOs.

Proper error handling is missing in some places, but I think it should be fine for a first version. I've also documented this in TODOs.

### DIFF
--- a/limitador/Cargo.toml
+++ b/limitador/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 
 # We make redis and infinispan optional to be able to compile for wasm32.
 [features]
-default = ["redis_storage", "infinispan_storage"]
+default = ["redis_storage"]
 redis_storage = ["redis", "r2d2", "tokio"]
 infinispan_storage = ["infinispan", "reqwest"]
 
--- a/limitador/Cargo.toml
+++ b/limitador/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 [features]
 default = ["redis_storage"]
 redis_storage = ["redis", "r2d2", "tokio"]
-infinispan_storage = ["infinispan", "reqwest"]
+infinispan_storage = ["infinispan", "reqwest", "openssl"]
 
 [dependencies]
 ttl_cache = "0.5"
@@ -32,6 +32,9 @@ r2d2 = { version = "0.8", optional = true }
 tokio = { version = "0.2", optional = true, features = ["rt-core", "macros", "time"] }
 infinispan = { version = "0.1", optional = true }
 reqwest = { version = "0.10", optional = true }
+
+# reqwest dependency. Needed here to avoid problems when compiling
+openssl = { version = "0.10.34", optional = true, features = ["vendored"] }
 
 [dev-dependencies]
 serial_test = "0.5"
--- a/limitador/Cargo.toml
+++ b/limitador/Cargo.toml
@@ -9,10 +9,11 @@ repository = "https://github.com/3scale-labs/limitador"
 documentation = "https://docs.rs/limitador"
 edition = "2018"
 
-# We make redis optional to be able to compile for wasm32.
+# We make redis and infinispan optional to be able to compile for wasm32.
 [features]
-default = ["redis_storage"]
+default = ["redis_storage", "infinispan_storage"]
 redis_storage = ["redis", "r2d2", "tokio"]
+infinispan_storage = ["infinispan", "reqwest"]
 
 [dependencies]
 ttl_cache = "0.5"
@@ -24,13 +25,13 @@ async-trait = "0.1.41"
 cfg-if = "1.0"
 prometheus = "0.10"
 lazy_static = "1.4"
-infinispan = "0.1"
-reqwest = "0.10"
 
 # Optional dependencies
 redis = { version = "0.17", optional = true, features = ["connection-manager"] }
 r2d2 = { version = "0.8", optional = true }
 tokio = { version = "0.2", optional = true, features = ["rt-core", "macros", "time"] }
+infinispan = { version = "0.1", optional = true }
+reqwest = { version = "0.10", optional = true }
 
 [dev-dependencies]
 serial_test = "0.5"
--- a/limitador/Cargo.toml
+++ b/limitador/Cargo.toml
@@ -24,6 +24,8 @@ async-trait = "0.1.41"
 cfg-if = "1.0"
 prometheus = "0.10"
 lazy_static = "1.4"
+infinispan = "0.1"
+reqwest = "0.10"
 
 # Optional dependencies
 redis = { version = "0.17", optional = true, features = ["connection-manager"] }
--- a/limitador/src/storage/infinispan/counters.rs
+++ b/limitador/src/storage/infinispan/counters.rs
@@ -1,0 +1,124 @@
+// Infinispan does not support counters with TTL. This module provides some
+// functions on top of the Infinispan client to be able to work with counters
+// with TTLs.
+//
+// Counters in Infinispan don't have a TTL. So to implement a counter, apart
+// from the actual counter, we need a helper key with a TTL that simply tells us
+// whether the counter is still valid.
+//
+// This is how it works in more detail: When a counter is created, apart from
+// the actual Infinispan counter, this module creates a regular entry with the
+// same name as the counter and with a TTL. The TTL is set to the duration of
+// the limit. So, for a limit of 10 requests per minute, the TTL would be a
+// minute. While the key has not expired, we know that the value of the counter
+// is valid. When the key has expired, the value of the counter no longer
+// applies.
+
+use crate::storage::infinispan::response::response_to_string;
+use infinispan::errors::InfinispanError;
+use infinispan::{request, Infinispan};
+use std::time::Duration;
+
+pub struct CounterOpts {
+    initial_value: i64,
+    ttl: Duration,
+}
+
+impl CounterOpts {
+    pub fn new(initial_value: i64, ttl: Duration) -> Self {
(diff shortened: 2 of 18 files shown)


limitador/src/storage/infinispan/counters.rs
```diff
@@ -0,0 +1,124 @@
...
+// is valid. When the key has expired, the value of the counter no longer
+// applies.
+
+use crate::storage::infinispan::response::response_to_string;
+use infinispan::errors::InfinispanError;
+use infinispan::{request, Infinispan};
+use std::time::Duration;
+
+pub struct CounterOpts {
+    initial_value: i64,
```
davidor, line 23: @unleashed This should be `CounterVal`, but when I was writing this I realized that it's not exposed in `infinispan-rs`.

limitador/src/storage/infinispan/sets.rs
```diff
@@ -0,0 +1,114 @@
...
+pub async fn add(
+    infinispan: &Infinispan,
+    cache_name: impl Into<String>,
+    set_key: impl Into<String>,
+    element: impl Into<String>,
+) -> Result<(), StorageErr> {
+    let cache_name = cache_name.into();
+    let set_key = set_key.into();
+
+    dist_lock::lock(infinispan, &cache_name, &set_key).await?;
```
unleashed, line 42: This is now calling for the lock guard type :) - check out [`MutexGuard`](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html) for an example.

You don't need to make it so it guards _data_, but it can guard the obvious bugs that show up in this fn. In particular, when you are done with the fn you want to release the lock if you created it successfuly, but you have at least 3 exit points that won't release a taken lock. Perhaps the issue is transient network error, perhaps it's a permanent reset, but the guard will force the code to try to release the lock when leaving this context.

Same thing applies to `delete`.

limitador/Cargo.toml
```diff
@@ -9,10 +9,11 @@ repository = "https://github.com/3scale-labs/limitador"
 documentation = "https://docs.rs/limitador"
 edition = "2018"
 
-# We make redis optional to be able to compile for wasm32.
+# We make redis and infinispan optional to be able to compile for wasm32.
 [features]
-default = ["redis_storage"]
+default = ["redis_storage", "infinispan_storage"]
```
unleashed, line 14: I'd say make Redis the default and avoid compiling in Infinispan, because, well, is it expected that users will want to support both at the same time with a single service?

unleashed (changes requested):
Check out the comments

limitador/src/storage/infinispan/counters.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    if response.status() == 404 {
+        // TODO: we could use "reset" here, but it's not implemented in the
+        // client yet. So for now, delete and create.
+        let _ = infinispan
+            .run(&request::counters::delete(&counter_key))
+            .await?;
+
+        // TODO: the type of counter and its attributes should be configurable.
+        // For now let's use "weak" counters with default attributes.
```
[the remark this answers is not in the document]
davidor: The client does not allow to configure distributed caches yet, so this does not really matter at this point.
My objective was to first make it work on a single Infinispan server ("local cache") and then worry about the different distributed cache types Infinispan offers if we think we should continue with this.

limitador/Cargo.toml
davidor: The image of the service that we publish needs both, but you're right, the rest of users don't need both.

limitador/src/storage/infinispan/counters.rs
davidor: Similar to https://github.com/3scale-labs/limitador/pull/38#discussion_r629225082

limitador/src/storage/infinispan/sets.rs
davidor: I don't see how to use MutexGuard in this context.

unleashed: The fn can be something like:
```rust
pub async fn lock(client: ..., cache_name: &str, set_key: &str) -> Result<DistLockGuard> {
    // [...]
    Ok(DistLockGuard {
        client,
        set_key,
        // any other info needed to unlock
    })
}
```
Then `DistLockGuard` should have:
```rust
impl Drop for DistLockGuard {
    fn drop(&mut self) {
        // perform unlock here
    }
}
```
Since you are using async, if the type is `Pin`-ned at any point, you'd want to check out [how to drop it](https://doc.rust-lang.org/std/pin/index.html#drop-implementation).

You can also add an `unlock` fn in this type, and if you do so you should probably have a flag about the state of the lock and prevent it from being called concurrently from other threads, ie `!Sync` (`Send` looks ok unlike mutexes which usually require that the same thread performing the lock should do the unlock).

limitador/src/storage/infinispan/dist_lock.rs
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            if retries >= RETRIES {
+                return Err(StorageErr {
+                    msg: "can't acquire lock".into(),
+                });
+            }
+
+            retries += 1;
+
+            tokio::time::delay_for(RETRIES_INTERVAL).await;
```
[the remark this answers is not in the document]
davidor: I wouldn't worry about making those 2 constants configurable at this point considering that this is a first version.
I think the point about `DistLockGuard` is more important, but I haven't come up with a working solution for that yet.

limitador/src/storage/infinispan/sets.rs
```diff
@@ -0,0 +1,114 @@
...
+) -> Result<HashSet<String>, InfinispanError> {
+    let get_entry_response = infinispan
+        .run(&request::entries::get(cache_name, &set_key))
+        .await?;
+
+    let set = if get_entry_response.status() == 404 {
+        HashSet::new()
+    } else {
+        // TODO: handle other errors
+        serde_json::from_str(&response_to_string(get_entry_response).await).unwrap()
```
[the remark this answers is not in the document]
davidor: I think this can only be caused by a programming error. This fails when we try to fetch an entry from infinispan and its body cannot be parsed has a Set. This probably means that limitador wrote something incorrect in that key.

We could return an `InconsistentData` error or similar in a case like this, and deal with it in an upper layer, but I don't think we can do much about it other than exiting and showing a more detailed error explaining what caused the crash.

Something we should do, but not sure if we should get to that point in this first version. I'd rather give a higher priority in trying the distributed cache types offered by Infinispan at this point than dealing with these edge cases.

limitador/Cargo.toml
davidor: Done in new commit: https://github.com/3scale-labs/limitador/pull/38/commits/1460a7e87c1651f93af2c52a14efe9853cb0e805

davidor:
@unleashed I rebased on top of master to be able to build the docker image with rust 1.52. I realized that this was missing: https://github.com/3scale-labs/limitador/pull/38/commits/5caa75d71c80ec7c4dd0cb5507de6635042ca1fa
Also, I added https://github.com/3scale-labs/limitador/pull/38#discussion_r631180066

Nothing else has changed for now.

limitador/src/storage/infinispan/sets.rs
davidor: We can't easily make the Drop solution work for now without making some changes in the infinispan-rs lib first.
I've added a new commit (the last one in this PR) that solves the problem in a different way.

limitador/src/storage/infinispan/sets.rs
```diff
@@ -38,15 +38,48 @@ pub async fn add(
 ) -> Result<(), StorageErr> {
     let cache_name = cache_name.into();
     let set_key = set_key.into();
+    let element = element.into();
 
     dist_lock::lock(infinispan, &cache_name, &set_key).await?;
 
+    let res = add_set(infinispan, cache_name.clone(), set_key.clone(), element).await;
+
+    dist_lock::release(infinispan, &cache_name, &set_key).await?;
```
unleashed, line 47: :thinking: :thought_balloon: how can I distinguish whether the `add_set` failed or the `release` failed? I think we want to distinguish the cases, because `dist_lock::release`-failure can be a soft-fail with no retry needed, whereas `add_set`-failure would be much more like a real, problematic failure.

davidor: I think both need retries. The difference is that if after retrying `add_set` still fails, we should send the error to the user whereas with `release` we shouldn't. Actually, `release` does not even need to run on request time.
Given that we don't have the logic for retries yet, not even in the other redis implementations, I suggest we tackle this on a future PR.

unleashed: But adding to a set being successful might not need a retry (if only it might be noted down that the lock could not be successfully released), no?

unleashed: Ok with me to tackle it in a future PR.

unleashed (approved):
r+ from me when the last commit is properly titled and/or squashed nicely into history :)

PR #40, opened 2021-05-07
Server: allow to select Infinispan as the storage for the limits

davidor:
~This PR includes the commits from #38, so that one should be merged first. That's why I'm creating this one as a draft.~

This PR continues the work started in #38 . This PR uses the code introduced in the lib in #38 to allow users to select infinispan as the storage for the limits/counters when running the server. To do that, users simply need to set the `INFINISPAN_URL` env.

### DIFF
--- a/limitador-server/Cargo.toml
+++ b/limitador-server/Cargo.toml
@@ -10,12 +10,14 @@ edition = "2018"
 [dependencies]
 limitador = { path = "../limitador", features = ['infinispan_storage'] }
 tokio = { version = "0.2", features = ["full"] }
+thiserror = "1.0"
 tonic = "0.3"
 prost = "0.6"
 prost-types = "0.6"
 serde_yaml = "0.8"
 log = "0.4"
 env_logger = "0.8"
+url = "2.2"
 actix-web = "3"
 actix-rt = "1"
 paperclip = { version = "0.5", features = ["actix"] }
--- a/limitador-server/Cargo.toml
+++ b/limitador-server/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-limitador = { path = "../limitador" }
+limitador = { path = "../limitador", features = ['infinispan_storage'] }
 tokio = { version = "0.2", features = ["full"] }
 tonic = "0.3"
 prost = "0.6"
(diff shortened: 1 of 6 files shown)


limitador-server/Cargo.toml
```diff
@@ -16,6 +16,7 @@ prost-types = "0.6"
 serde_yaml = "0.8"
 log = "0.4"
 env_logger = "0.8"
+url = "2.2"
```
unleashed, line 19: Not using the 3scale-rs fork. I am disappoint.

limitador-server/src/main.rs
```diff
@@ -26,35 +30,48 @@ pub enum Limiter {
...
-                    rate_limiter_builder = rate_limiter_builder.with_prometheus_limit_name_labels()
-                }
+    fn storage_type() -> StorageType {
+        let redis_url = env::var(REDIS_URL_ENV);
+        let infinispan_url = env::var(INFINISPAN_URL_ENV);
 
-                Limiter::Blocking(rate_limiter_builder.build())
+        match (redis_url, infinispan_url) {
+            (Ok(_), Ok(_)) => {
+                panic!("Please set the Redis or the Infinispan URL, but not both");
```
unleashed, line 58: Panicking is not (user) error management - you should exit the process nicely, not with a screenful of internal info and a backtrace.

limitador-server/src/main.rs
```diff
@@ -26,35 +30,48 @@ pub enum Limiter {
...
             }
-        };
+            (Ok(url), Err(_)) => StorageType::Redis { url },
+            (Err(_), Ok(url)) => StorageType::Infinispan { url },
+            _ => StorageType::InMemory,
```
unleashed, line 62: How about using the awesome `Option::xor` for this one-expression-er?
```rust
env::var(REDIS_URL_ENV).map(|url| StorageType::Redis { url })
    .xor(env::var(INFINISPAN_URL_ENV).map(|url| StorageType::Infinispan { url }))
    .or(StorageType::InMemory)
```

limitador-server/src/main.rs
davidor: :+1:
It's using `thiserror` now.

limitador-server/src/main.rs
davidor: Decided not to do this because it changes the semantics. Doesn't detect when both urls have been set and defaults silently to use in-memory.

unleashed (approved):
bors r+

PR #44, opened 2021-06-01
Allow to choose an Infinispan cache for the limits instead of creating one

davidor:
This PR makes changes both in the lib and the server to allow users to choose the Infinispan cache that they'd like to use. In the server side there's a new ENV, `INFINISPAN_CACHE_NAME` for this. In the lib, I added a new builder for the `InfinispanStorage` struct that accepts a cache name.

### DIFF
--- a/limitador-server/docs/configuration.md
+++ b/limitador-server/docs/configuration.md
@@ -31,6 +31,16 @@ variables:
 - Format: integer.
 
 
+## INFINISPAN_CACHE_NAME
+
+- The name of the Infinispan cache that Limitador will use to store limits and
+counters. This variable applies only when [INFINISPAN_URL](#infinispan_url) is
+set.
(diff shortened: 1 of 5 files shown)


limitador-server/docs/configuration.md
```diff
@@ -31,6 +31,16 @@ variables:
 - Format: integer.
 
 
+## INFINISPAN_CACHE_NAME
```
unleashed, line 34: The amount of environment variables is starting to get out of hand... :)

Yes, this is intended to be run with its own container image, but perhaps it would be good to look into also accepting a configuration file providing fallback settings for missing envs?

unleashed (approved):
LGTM - check out comments

limitador-server/docs/configuration.md
davidor: Yes, we should consider that. There are more config params coming :D

PR #50, opened 2021-10-06
nightly tests

eguzki:
Rust toolkit is not pinned to an specific version. Thus, latest lint tools checks are being tested. As some new checks may start failing, the nightly cron job will report if they fail.


.github/workflows/rust.yml
```diff
@@ -7,6 +8,8 @@ on:
   pull_request:
     branches:
       - '*'
+  schedule:
+    - cron: '5 4 * * 6'
```
andrewdavidmackenzie, line 12: That will only run on Sundays (day 6) no?
Don't you want '*' in the last field to run every night?
you chose 5 past 4PM for some particular reason? :-)
https://jasonet.co/posts/scheduled-actions/
e.g. midnight every day would be:
0 0 * * *

andrewdavidmackenzie (changes requested):
Review cron format for GH Actions as I don-t think this is what you intended?

PR #51, opened 2021-12-22
Perform semver-compatible dependency updates and move to edition 2021

unleashed:
Moving to the new edition essentially just moves to a new lock file format since we are not using any of the changed features, and the updated dependencies are semver-compatible with the specs.


rahulanand16nov (approved):
LGTM

PR #54, opened 2022-01-21
Update Limitador to Tokio 1 async reactor

unleashed:
This PR updates the whole dependency tree to move to Tokio 1 at once.

Some dependency black magic has had to happen to ensure everything is on the right side of the runtime. It uses a pre-release, beta version of Actix 4 and a fork of paperclip to adapt to it, as well as newer, tokio1-compatible versions of multiple dependencies.

Because of the above we should seek some validation that the software runs in production just fine with no regressions (and hopefully better performance!).

While at it this also enables clippy lints as errors everywhere and adds minimal crate metadata.


eguzki (approved):
LGTM

PR #55, opened 2022-01-25
server/envoy_rls: re-enable clippy reducing exception scope to proto generated config module

unleashed:
We were disabling clippy completely under envoy_rls, which was a way bigger gun than needed to silence a warning we cannot do anything about. This limits the scope to the relevant module and relevant lint.


rahulanand16nov (approved):
I was thinking, should we even care about Clippy errors from content on which we have zero control? Regardless, LGTM :)

unleashed:
We don't have direct control over the generation process, although we don't autogenerate it but instead we commit the results (so we have control, but no one wants to keep editing those files to then deal with the diffs).

The important thing is that clippy lints about many things and shutting it down completely is not great, because there is currently no other breakage and there are lints for stuff that indicate bugs or inefficiencies, not just style. If we see regressions like those, either there is a problem with the source definitions or, most likely, with the tool, and in both cases we should look into reporting them rather than silently taking them.

PR #56, opened 2022-01-27
Update protobuf definitions 20220127

unleashed:
This PR updates protobuffer definitions for the gRPC service using Envoy's RLS. Unfortunately this pulls in a lot of files that we don't really need, but what is/isn't needed is subject to change and requires a pruning process that has to be repeated each time we update. So we just pull in all proto definitions from the relevant repositories (documented in the `README.md` file under `vendor/protobufs` along with the reason this is being done) and hope the whole package won't be too big.


eguzki:
LGTM

eguzki (approved):

PR #57, opened 2022-01-31
limitador-server: release 0.5.0

unleashed:
This is a 0.4 backwards-compatible release, but bumped because internals have changed quite a bit regarding the async reactor and performance characteristics _might_ differ. Thoroughly testing with heavy workloads is recommended before placing in production.

CHANGELOG excerpt:

> ## 0.5.0 - 2022-01-31
> 
> ### Added
> 
> - Support for Infinispan (tested with 11.0.9.Final). You can enable it via the `INFINISPAN_URL` env variable.
> 
> ### Changed
> 
> - Moved to the Tokio 1 async reactor.

There are other internal changes that don't impact users, namely updating the RLS protocol definition and some refactorings to explicitly expose fallible operations for future development.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,9 +21,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.0.0-beta.18"
+version = "3.0.0-beta.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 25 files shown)


unleashed:
Yeah, I suspect it is due to the deployment pointing to a non-existing image (0.5.0) and timing out.

PR #61, opened 2022-04-13
Only use key utility functions when needed

alexsnaps:
The utility functions for key manipulation are only used with an
external storage is used. This removes warning when using only in memory
storage, e.g. when building targeting wasm32

Minor thing, but removes warnings... which is always nice 😃


rahulanand16nov (approved):
I haven't explored Limitador's codebase much but a quick peak suggests it's right :+1:

PR #65, opened 2022-05-25
bump alpine image

MStokluska:


eguzki (approved):

unleashed:
#58 does this along fixing thr build with 3.15. We should probably update that PR.

MStokluska:
> #58 does this along fixing the build with 3.15. We should probably update that PR.

@unleashed just noticed your previous PR too, would it make sense to bump alpine to 3.16 instead of 3.15?

unleashed:
> > #58 does this along fixing the build with 3.15. We should probably update that PR.
> 
> @unleashed just noticed your previous PR too, would it make sense to bump alpine to 3.16 instead of 3.15?

Definitely.

eguzki:
merged #58, please rebase and update both builder and runtime image

PR #68, opened 2022-05-25
v0.5.1 release

eguzki:


eguzki:
The test failing is because the image tag has not been created yet :)

PR #73, opened 2022-05-31
Ignore authority header check when it contains `|` char

### LINKED ISSUE #53: Request with `|` (pipe) character in authority header is auto-rejected by Limitador
This issue describes the finding of auto-rejection of requests that contains the pipe character in their `:authority` header field. I tracked down this problem to a dependency named [h2](https://github.com/hyperium/h2).

Following was from the trace logs:
```
malformed headers: malformed authority (b"outbound|8081||limitador.rahul-test.svc.cluster.local"): invalid uri character
```

**Why there is a pipe character in the authority header?** 
That format should be familiar if you have seen Istio's cluster config. Still, if not, that format is auto-generated by Istio's control plane by looking at its service registry.  So, when you add a Limitador service in the cluster where Istio is also present, the service registry picks up the service and sends that to all the Envoys present in the cluster (gateway or sidecar) as a [cluster](https://www.envoyproxy.io/docs/envoy/latest/api-v3/clusters/clusters). And the name of that cluster is of that format as seen above in the log.

As Limitador is used right now, it receives a callout request from an [Envoy's RateLimit filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rate_limit_filter) and one particular behavior of envoy is to use the cluster name as `:authority` header of the callout. Hence why you get the pipe character in the header.

**Potential way to solve this problem?**
I have seen a few comments around updating dependencies but I cannot find them right now. One option is to update the dependency which is already a bit old.

Second, I am going to create an issue in the Istio and Envoy repo to highlight this behavior to them. Possibly Istio can change the cluster format or Envoy can allow overriding authority?

cc @maleck13 @unleashed

rahulanand16nov:
This PR fixes #53 by forking https://github.com/hyperium/h2 and allowing requests with authority header containing pipe `|` characters.

The fix is present at: https://github.com/rahulanand16nov/h2/commit/d956cec4ac9dc1ee94aac416f5d338d5b296f902

It's not a security risk since the authority header is used nowhere in the logic of limitador (correct me if I am wrong). A similar issue will be created in the dependencies repo to fix it.

An image that contains this fix and works: `quay.io/rahanand/limitador:test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -831,8 +831,7 @@ dependencies = [
 [[package]]
 name = "h2"
 version = "0.3.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9f1f717ddc7b2ba36df7e871fd88db79326551d3d6f1fc406fbfd28b582ff8e"
+source = "git+https://github.com/rahulanand16nov/h2?branch=v0.3.11-pipe#d956cec4ac9dc1ee94aac416f5d338d5b296f902"
 dependencies = [
  "bytes",
  "fnv",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,4 +7,4 @@ lto = true
 codegen-units = 1
 
 [patch.crates-io]
-h2 = { git = "https://github.com/rahulanand16nov/h2", branch = "v0.3.11-pipe" }
+h2 = { git = "https://github.com/rahulanand16nov/h2", branch = "v0.3.11-pipe" } # This patch can be removed once hyperium/h2#619 is closed.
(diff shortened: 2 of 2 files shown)


rahulanand16nov:
I have created an issue in the upstream so we can remove our fork in the future when it's fixed. https://github.com/hyperium/h2/issues/619

alexsnaps:
So that's all hardcoded afaict: https://github.com/istio/istio/blob/32afc27ea91ef68e4e5f0ee3dd47d0aedfba35f1/pilot/pkg/model/service.go#L691-L693

rahulanand16nov:
Yup, can close now but the issue should remain open.

PR #76, opened 2022-06-08
Impl `From<&str> for Namespace` instead of `FromStr` as it can't fail

alexsnaps:
`Namespace` implements `FromStr` from the [stdlib](https://doc.rust-lang.org/std/str/trait.FromStr.html). But afaict `Namespace` _is_ a `String`, as such any `String` represents a valid `Namespace`. `FromStr` can fail tho. Even if our implementation had `type Err = core::convert::Infallible` it still means you have to call `unwrap()` on the `Result` from the the `try_into()` call.

There might have been a reason for this, but given the existing infallible `From<String>` implementation, I can't see it.

This changes this to `impl From<&str> for Namespace`, just as it already did for `From<String>` and makes the `.into() -> Namespace` removing the intermediary `Result` that was always a success anyways. As such it removes a bunch of `.unwrap()` from the code base. 

Part one of more to come...

### DIFF
--- a/limitador-server/src/http_api/request_types.rs
+++ b/limitador-server/src/http_api/request_types.rs
@@ -41,7 +41,7 @@ impl From<&LimitadorLimit> for Limit {
 impl From<Limit> for LimitadorLimit {
     fn from(limit: Limit) -> Self {
         let mut limitador_limit = Self::new(
-            limit.namespace.as_str(),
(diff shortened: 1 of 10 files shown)


rahulanand16nov (approved):
Always good to reduce `unwrap`s :pray:

PR #77, opened 2022-06-08
`RaterLimiter` to take a `&Namespace` 

alexsnaps:
This is based of #76 
I think this is cleaner (and avoids a couple `.clone()`s along the way). If we all agree, this could be merged.


rahulanand16nov (approved):
LGTM

Issue #74, opened 2022-06-02
Redefine authority of `Limit`s


alexsnaps:
Currently `Limit`s are "owned" by Limitador. It gets them from a file and then the set of `Limit` can be altered using the REST end points. In a k8s world, that file is a mounted `ConfigMap` and the end-point is invoked based on the changes made to the Custom Resources, i.e. `RateLimitPolicy`'ies. All these `Limit`s are persisted into Redis, when Limitador is configured to use it. 

The problem with this approach is that we end up with two authorities over `Limit`s, that all have a different lifecycle. e.g. Limitador is up & running, loaded an initial set of `Limit`s from the `ConfigMap` as it was when it started. If that `ConfigMap` is changed, Limitador will only pick up those changes when restarted. In the meantime, some `RateLimitPolicy`'ies are being added and the operator changes the config using the REST endpoints on Limitador. For some reason Limitador is down and gets restarted. Currently, only the `Limit`s from the (whatever state it is in) `ConfigMap` are being kept. So for the `Counter`s associated with them. All the other `Limit`s are discarded. 

### Fixing this

While we could "simply" make the operator smarter and "work around" these shortcomings, some of us believe that clearly defining a _single authoritative source_ for the `Limit`s would make the system simpler to manage. After discussing this with @eguzki , we thought a forward would be to _not_ persist `Limit`s in Redis anymore but to only keep them in-memory. They'd be loaded from a single config file from "disk" at start up, and changes to that file would be picked up by Limitador without the need for a restart. In a k8s world, the control plane would be responsible for updating that file accordingly, based of changes to Custom Resources and other `ConfigMap` it'd care about. That way the lifecycle of the different components (the k8s cluster(s), Limitador & Redis) become decoupled & changes are applied consistently.

#### A word on sharding

Interestingly this comes with an additional advantage. If we need to shard the workload across multiple Limitador instances, how that sharding happens is now entirely controlled by the control plane: how to configure Limitador instances and how gateway(s) get to query those. 

### Todo

- [x] #78 
- [x] Load the Limit from the file at start up (and not reconcile with the on-disk state, as they'd be none)
- [x] #79 
- [x] Kuadrant/limitador-operator#21

rahulanand16nov:
Just to clarify, if the operator is down and comes back then it will need to reconcile the current state of the `RateLimits` with the current state of `ConfigMap` instead of sending new limits via an HTTP endpoint.

didierofrivia:
[...]

PR #78, opened 2022-06-13
Keeps all Limits in memory only

alexsnaps:
Part of addressing #74


eguzki (commented):
small doc fix should be done https://github.com/Kuadrant/limitador/blob/25bb7762192324b61cbacc73d19c21682b45ea8e/limitador/src/lib.rs#L5

limitador/src/storage/mod.rs
```diff
@@ -21,38 +23,230 @@ pub enum Authorization<'c> {
...
-    fn get_limits(&self, namespace: &Namespace) -> Result<HashSet<Limit>, StorageErr>;
-    fn delete_limit(&self, limit: &Limit) -> Result<(), StorageErr>;
-    fn delete_limits(&self, namespace: &Namespace) -> Result<(), StorageErr>;
+pub struct Storage {
+    limits: RwLock<HashMap<Namespace, HashSet<Limit>>>,
+    counters: Box<dyn CounterStorage>,
+}
+
+pub struct AsyncStorage {
+    limits: RwLock<HashMap<Namespace, HashSet<Limit>>>,
```
eguzki, line 32: Nice. HashSet enforces uniqueness for limits, doesn't it?

eguzki (approved):
Looking super good. 

I failed to understand every bit, but I got the main idea. Limits live in the `Storage` struct (not a trait any more) and counters accessor is a trait (well, a smart pointer to an object implementing `CounterStorage`). Redis, InMemmory and Infinispam all implement `CounterStorage`

limitador/src/storage/mod.rs
```diff
@@ -21,38 +23,230 @@ pub enum Authorization<'c> {
...
+    }
+}
+
+impl Default for Storage {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl AsyncStorage {
```
eguzki, line 130: Just for curiosity. Does it make sense to keep both async and sync storage version? If I am not wrong it is all about using the redis sync or async client. Shouldn't we choose one and stick to it?

limitador/src/storage/mod.rs
```diff
@@ -21,38 +23,230 @@ pub enum Authorization<'c> {
...
+            None => HashSet::new(),
+        }
+    }
+
+    pub fn delete_limit(&self, limit: &Limit) -> Result<(), StorageErr> {
+        let mut limits = HashSet::new();
+        limits.insert(limit.clone());
+        self.counters.delete_counters(limits)?;
+
+        let mut limits = self.limits.write().unwrap();
```
rahulanand16nov, line 77: I am always a little hesitant with so many unwrap but I understand likelyhood is low. Do you think we should manage results instead of panicing. Maybe I am being too conservative.

limitador/src/storage/mod.rs
```diff
@@ -21,38 +23,230 @@ pub enum Authorization<'c> {
...
+
+        if let Some(counters_by_limit) = limits.get_mut(limit.namespace()) {
```
rahulanand16nov, line 79: `counters_by_limit` Not sure about what that name means or if it's correct?

rahulanand16nov (commented):
Looks good and added my thoughts. I am not able to review redis related parts yet but will do in the future as I get more accustomed to the code.

limitador/src/storage/mod.rs
rahulanand16nov: @alexsnaps Looks like we including the limit's name into the hash for a Limit. I think two limits with everything same except name should be same as well right?

alexsnaps: @eguzki yes… but has it did before, as `Limit` was used as the key in a `HashMap`.
@rahulanand16nov I didn't change anything there really… but I tend to agree with you indeed. `name` doesn't semantically mean anything to the `Limit`, otoh the `max_value` is part of the `PartialEq#eq` impl. which I think is wrong. I'd rather _not_ conflate these changes with this PR tho. *But* I think we should address these issues too!

limitador/src/storage/mod.rs
alexsnaps: Well… again, nothing new here. But these `.unwrap()` are on `LockResult<RwLockReadGuard>` which would return an `Err` if the lock is "poisoned". That happens when a thread panic'ed while holding an exclusive lock. This indicates that whatever this lock was guarding probably is "corrupted". But so that would mean: something went real bad in the past, for which we didn't cleaning recover from _and_ some other thread is now trying to use that (possibly) corrupted state. 
Until we find a strategy to recover from a poisoned state, I think panic'ing completely is probably the best option anyhow. One way would be to explicitly table flip and `process::exit(1)` on such a situation… unsure how much I want to litter the code with these conditions tho. 

Again, if we want to change that (which I'm certainly open to, if only to start the discussion on how to achieve that), I wouldn't include these functional changes in this PR.

limitador/src/storage/mod.rs
alexsnaps: No it's not... left over from the code I moved… I'll fix! Good catch thanks!

limitador/src/storage/mod.rs
```diff
@@ -21,38 +23,230 @@ pub enum Authorization<'c> {
...
     fn is_within_limits(&self, counter: &Counter, delta: i64) -> Result<bool, StorageErr>;
     fn update_counter(&self, counter: &Counter, delta: i64) -> Result<(), StorageErr>;
     fn check_and_update<'c>(
         &self,
         counters: &HashSet<&'c Counter>,
         delta: i64,
     ) -> Result<Authorization<'c>, StorageErr>;
-    fn get_counters(&self, namespace: &Namespace) -> Result<HashSet<Counter>, StorageErr>;
+    fn get_counters(&self, limits: HashSet<Limit>) -> Result<HashSet<Counter>, StorageErr>;
+    fn delete_counters(&self, limits: HashSet<Limit>) -> Result<(), StorageErr>;
```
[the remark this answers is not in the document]
alexsnaps: Kept method signatures as is to avoid making too many changes. 
I have another PR coming that changes that very trait.

alexsnaps:
> Before merging this, I would tag the current HEAD of `main` branch. Optionally create a branch to support existing deployments.
> 
> Alternatively, we could use the existing `v0.5.1` tag

I'd rather use the the `v0.5.1` or whatever version users were using _if_ we need to fork. And possibly cherry-pick further changes along the way. But I might be overly conservative

limitador/src/storage/mod.rs
eguzki: ok, got it... I think. 

Thanks for the explanation

PR #79, opened 2022-06-15
Monitor config file for changes and reload when needed

alexsnaps:
See issue #74 

### What has changed

 - bafcb99 Adds the behavior to reload the limit file off disk when it changed and do the changes
 - 4394736 Returns proper errors when handling the limit file off disk

### Testing it locally

1. Build locally
`$ cargo build`
2. Start limitador server, using the example config
`$ LIMITS_FILE=limitador-server/examples/limits.yaml RUST_LOG=info ./target/debug/limitador-server`
3. Hit the HTTP end point to list limits at:
http://localhost:8080/limits/test_namespace
4. Change a value in the config file at `limitador-server/examples/limits.yaml` and save
5. Reload the URL and the changes should be visible
6. An invalid config file would trigger an error to be logged and be ignored (i.e. the old config remains active)

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -10,11 +10,15 @@ use limitador::storage::infinispan::{Consistency, InfinispanStorageBuilder};
 use limitador::storage::redis::{AsyncRedisStorage, CachedRedisStorage, CachedRedisStorageBuilder};
 use limitador::storage::{AsyncCounterStorage, AsyncStorage};
 use limitador::{AsyncRateLimiter, AsyncRateLimiterBuilder, RateLimiter, RateLimiterBuilder};
+use notify::event::{DataChange, ModifyKind};
+use notify::{Error, Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use std::convert::TryInto;
+use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 use std::{env, process};
 use thiserror::Error;
+use tokio::runtime::Handle;
 use url::Url;
 
 mod envoy_rls;
@@ -55,8 +59,6 @@ impl Limiter {
             StorageType::Infinispan { url } => Self::infinispan_limiter(&url).await,
             StorageType::InMemory => Self::in_memory_limiter(),
         };
-
-        rate_limiter.load_limits_from_file().await;
 
         Ok(rate_limiter)
     }
@@ -194,15 +196,13 @@ impl Limiter {
         }
     }
 
-    async fn load_limits_from_file(&self) {
-        if let Ok(limits_file_path) = env::var(LIMITS_FILE_ENV) {
-            let f = std::fs::File::open(limits_file_path).unwrap();
-            let limits: Vec<Limit> = serde_yaml::from_reader(f).unwrap();
+    pub async fn load_limits_from_file<P: AsRef<Path>>(&self, path: &P) {
+        let f = std::fs::File::open(path).unwrap();
+        let limits: Vec<Limit> = serde_yaml::from_reader(f).unwrap();
 
-            match &self {
-                Self::Blocking(limiter) => limiter.configure_with(limits).unwrap(),
-                Self::Async(limiter) => limiter.configure_with(limits).await.unwrap(),
-            }
+        match &self {
+            Self::Blocking(limiter) => limiter.configure_with(limits).unwrap(),
+            Self::Async(limiter) => limiter.configure_with(limits).await.unwrap(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


limitador-server/src/main.rs
```diff
@@ -219,6 +249,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
     };
 
+    let _watcher = if let Ok(limits_file_path) = env::var(LIMITS_FILE_ENV) {
```
eguzki, line 252: nothing to be done with the watcher? maybe register a signal to stop watching?

eguzki (approved):
owesome!

limitador-server/src/main.rs
```diff
@@ -219,6 +249,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        if let Err(e) = rate_limiter.load_limits_from_file(&limits_file_path).await {
+            eprintln!("Failed to load limit file: {}", e);
+            process::exit(1)
+        }
+
+        let limiter = Arc::clone(&rate_limiter);
```
eguzki, line 258: I wonder if this is the same as 
```rust
let limiter = rate_limiter.clone();
```

limitador-server/src/main.rs
alexsnaps: Well that's what this does… RAII… i.e. we keep it around, and things get cleaned up upon `drop()`, e.g. in the case of linux, that'd be [here](https://github.com/notify-rs/notify/blob/main/src/inotify.rs#L644-L650)

limitador-server/src/main.rs
alexsnaps: As per the [documentation](https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references), it is indeed:

```rust
// The two syntaxes below are equivalent.
let a = foo.clone();
let b = Arc::clone(&foo);
```

Dunno why I used one over the other… which ever you prefer ¯\_(ツ)_/¯

limitador-server/src/main.rs
alexsnaps: btw, this is very much like automatic `defer`al as you'd know it from golang… only you _cannot_ forget to tell it to release the resources!

limitador-server/src/main.rs
eguzki: :+1:  thanks for the doc ref

limitador-server/src/main.rs
eguzki: Ok, let me verify if I understood. 

When `main()` function run, the execution point is blocked at 

```rust
run_http_server(&http_api_address, rate_limiter.clone()).await?;
```

When the process gets killed (with a signal), the execution continues in `main` returning `Ok(())`. Then, `_watcher` which is an Option holding the watcher, gets out of scope and dropped. Notify lib will clean up at that moment. 

is that right?

eguzki:
Two more questions:

* The watcher is waiting for `EventKind::Modify` events. What happens when the process starts? The file is already there but not changed. Does the modify even trigger?
* If I do not get it wrong, the `LIMITS_FILE_ENV` env var is optional, and if it does not exist, the watcher will not start. Effectively, the limits will be empty forever (forever is the lifetime of the process). I wonder if we need to enforce the existence of the env var and if not, exit the process with an error.

limitador-server/src/main.rs
alexsnaps: Yes that's exactly it. Or to be real pedantic, while you hit the return, the function doesn't return until `.drop()` has itself returned on the `Option` (that if it owns the `Watcher` eventually delegates to its `drop()`)… 

The [doc](https://doc.rust-lang.org/std/ops/trait.Drop.html) might provide more clarity than my explanation here… otherwise (and I recommend it in all cases) [Rust for Rustaceans](https://nostarch.com/rust-rustaceans) is excellent, and the [free first chapter](https://nostarch.com/download/samples/RustforRustaceans_Ch1.pdf) has the part on `Drop` and drop order on page 9!

alexsnaps:
> Two more questions:
> 
>     * The watcher is waiting for `EventKind::Modify` events. What happens when the process starts? The file is already there but not changed. Does the modify even trigger?

No the event triggers only when observed.

>     * If I do not get it wrong, the `LIMITS_FILE_ENV` env var is optional, and if it does not exist, the watcher will not start. Effectively, the limits will be empty forever (forever is the lifetime of the process). I wonder if we need to enforce the existence of the env var and if not, exit the process with an error.

That's the `Option` is for. If you did _not_ provide a `LIMITS_FILE` env var then yes… you effectively have a useless limitador instance 🤔 So I think you might be right there! 

Do you want to do this as part of this PR? I have one question about this tho: should we have a default? If so which? 
And now I really start hating the idea of a mandatory env variable and would much rather provide this as an argument to the process… wdyt?

eguzki:
| Do you want to do this as part of this PR? I have one question about this tho: should we have a default? If so which?
And now I really start hating the idea of a mandatory env variable and would much rather provide this as an argument to the process… wdyt?

I would say a different PR.

For the remaining questions: I agree that and env var should always be optional. The main question is, do we want to have this limits file path as a required input to the process? If yes, then I would go to command line arg. The libraries for manage command line args usually have clear mechanisms to flag some command line arg as `required`. It could even be a positional param instead of a optional param (e.g. `--limits-file`).
On the other hand, if limitador defines a default file as the source of limits (for example `/opt/limitador/etc/limits.yaml`), then I really do not care about being via env var or optional param (e.g. `--limits-file`) the mechanism to override the default one.

alexsnaps:
Looks like we have a similar view on this :)
Created #83 to track this… i.e. make a decision and impl. it.

PR #80, opened 2022-06-16
Simpler trait

alexsnaps:
### Some background 

When you look at the lifecycle of the data (i.e. how the data "flows" within limitador) there aren't much reasons to use references much. By moving ownership down into the lib, in many cases we can get close to no copy (and there is still room for improvement, e.g. the `Map<String, String>` for conditions & vars, but that's a topic for another day).

 - 8b6ca3e make the previous change somewhat "lighter". Initially within 692ff34, I was keeping a (Boxed, to keep the enum small) `Counter` in the `Authorization::Limited`. But then seeing that we don't need it, I changed it to only have the name of the limit that tripped. We _could_ still keep the whole `Counter` in there (I sort of like the API of `Authorization::Limited` having the counter that tripped). In which case the flow in the first change (692ff34) would be iterate while taking ownership of all the counters [as here](https://github.com/Kuadrant/limitador/pull/80/commits/692ff34c63b2d4ea2d95328d9ef13421c68e4b54#diff-ebd607498f8416f1e94e7592efd8580d8dd6798df308dee8f7f7f7f78fe73c18R40-R53), unless we trip a counter in which case we early return by passing ownership of said `Counter` to `Authorization::Limited`. 
 - d078d85 Uses a reference to a `HashSet<Limit>` which avoid having to clone all the `Limit` of one `HashSet` to merely build the same container. This addresses [a comment made on previous PR](https://github.com/Kuadrant/limitador/pull/78#discussion_r898977195) by @rahulanand16nov that rightfully highlighted that we could do better (and avoiding `clone()`ing that much). 

I hope this makes "some" sense.

### DIFF
--- a/limitador/src/lib.rs
+++ b/limitador/src/lib.rs
@@ -384,16 +384,16 @@ impl RateLimiter {
 
         let check_result = self
             .storage
-            .check_and_update(&counters.iter().collect(), delta)?;
+            .check_and_update(counters.into_iter().collect(), delta)?;
 
         match check_result {
             Authorization::Ok => {
                 self.prometheus_metrics.incr_authorized_calls(namespace);
                 Ok(false)
             }
-            Authorization::Limited(c) => {
+            Authorization::Limited(name) => {
                 self.prometheus_metrics
-                    .incr_limited_calls(namespace, c.limit().name());
+                    .incr_limited_calls(namespace, name.as_deref());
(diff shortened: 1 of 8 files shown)


eguzki (approved):

rahulanand16nov (approved):
Sorry for the late review, I was on PTO but LGTM.

PR #82, opened 2022-06-16
Bench improvements

alexsnaps:
I changed 3 things really:

 - e2dda39 removes use of deprecated Criterion APIs
 - b267e26 enables running bench with --no-default-features (i.e. only with in-memory counter storage)
 - 034adf1 since we don't want to hit limits, changed them to use `i64::MAX` (why is this a `i64`?!)

### DIFF
--- a/limitador/benches/bench.rs
+++ b/limitador/benches/bench.rs
@@ -1,4 +1,4 @@
-use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
+use criterion::{black_box, criterion_group, criterion_main, Bencher, BenchmarkId, Criterion};
 use rand::seq::SliceRandom;
 
 use limitador::limit::Limit;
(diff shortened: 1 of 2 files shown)


limitador/benches/bench.rs
```diff
@@ -54,70 +52,77 @@ struct TestCallParams {
...
-        &format!("check_rate_limited_and_update (redis)"),
-        |b: &mut Bencher, test_scenario: &&TestScenario| {
-            let storage = Box::new(RedisStorage::default());
-            bench_check_rate_limited_and_update(b, test_scenario, storage);
-        },
-        TEST_SCENARIOS.to_vec(),
-    );
+#[cfg(feature = "redis")]
+fn bench_redis(c: &mut Criterion) {
+    let mut group = c.benchmark_group("In memory");
```
eguzki, line 98: shouldn't it be `c.benchmark_group("redis");`

limitador/benches/bench.rs
```diff
@@ -54,70 +52,77 @@ struct TestCallParams {
...
-    c.bench_function_over_inputs(
-        &format!("check_rate_limited_and_update (redis)"),
-        |b: &mut Bencher, test_scenario: &&TestScenario| {
-            let storage = Box::new(RedisStorage::default());
-            bench_check_rate_limited_and_update(b, test_scenario, storage);
-        },
-        TEST_SCENARIOS.to_vec(),
-    );
+#[cfg(feature = "redis")]
+fn bench_redis(c: &mut Criterion) {
```
eguzki, line 97: `bench_redis` is pretty much the same as `bench_in_mem` except for the storage. Isn't there a way to generalize into

```rust
fn bench_generic(c: &mut Criterion, storage: Box<dyn CounterStorage>)
```

?

eguzki (approved):
Few comments added. Overall looks good!

limitador/benches/bench.rs
alexsnaps: @eguzki I actually did something like that (or almost, see below), but thought it might be less readable… So decided to drop the change. I'll let you judge of which version is better, see this change set: dc79ae5

Let me know whether to keep it (I'd squash it before merging) or whether to just drop it…

eguzki: It is ok as it is. It is a tests and readability matters.

Issue #81, opened 2022-06-16
Consolidate env variable names


alexsnaps:
Currently env variables are used to configure `limitador-server`. We should consolidate these in some sort of prefix/namespace. `LIMITADOR_` maybe? So that e.g. the currently `LIMITS_FILE` would become `LIMITADOR_LIMITS_FILE`.
We should also modify the logger to use the same prefix (currently `RUST_LOG` and `RUST_LOG_STYLE`)

<details>

### Config the logger

```diff
diff --git a/limitador-server/src/main.rs b/limitador-server/src/main.rs
index 280cdf4..110cf02 100644
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -5,6 +5,7 @@ extern crate log;
 
 use crate::envoy_rls::server::run_envoy_rls_server;
 use crate::http_api::server::run_http_server;
+use env_logger::Env;
 use limitador::errors::LimitadorError;
 use limitador::limit::Limit;
 use limitador::storage::infinispan::{Consistency, InfinispanStorageBuilder};
@@ -239,7 +240,13 @@ impl Limiter {
 
 #[actix_rt::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    env_logger::init();
+    if let Err(e) = env_logger::try_init_from_env(
+        Env::new()
+            .filter("LIMITADOR_LOG")
+            .write_style("LIMITADOR_LOG_STYLE"),
+    ) {
+        eprintln!("Couldn't initializer logger: {}", e);
+    }
 
     let rate_limiter: Arc<Limiter> = match Limiter::new().await {
         Ok(limiter) => Arc::new(limiter),

```

</details>

alexsnaps:
Might be "overruled" by #83 if we go for command line args… Or should we keep the logging configuration thru env vars maybe?

alexsnaps:
I think with the work done as part of CLI #92 We can leave this as is (so things remain backwards compatible, yet the new CLI based approached is favored)

PR #84, opened 2022-06-21
Moved all env var configuration into one place

alexsnaps:
Moves all environment variable parsing into one location, the `Configuration` struct, for now only initialized from `from_env()` variables, but opens the seam for more options (e.g. command line args and/or config file)

Prep work for #81 and adding cmd line args?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -532,9 +532,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-utils"
-version = "0.8.9"
+version = "0.8.10"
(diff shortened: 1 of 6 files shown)


eguzki:
> Prep work for https://github.com/Kuadrant/limitador/issues/81 and adding cmd line args?

Looking forward to it

eguzki (approved):
Code looks great. Much clearer and structured that is used to be. 

Great refactor for next steps preparation

alexsnaps:
Resolved conflicts in `Cargo.lock`

PR #89, opened 2022-07-13
We need to insert a HashSet before doing anything

alexsnaps:
Stupid bug… yet a bug.
@eguzki can you confirm this fixes this issue? If so, I'll merge!
Thanks 🙏


eguzki:
hitting two limits does no longer break it. :muscle:

eguzki (approved):

eguzki:
maybe some testing :?

PR #91, opened 2022-07-13
inotify on linux sends a different type

alexsnaps:
I'd love if you could test this @eguzki - but based on the code I read, this should be fine on linux too now


eguzki:
Now, I can see log line about "Reloaded limit file". However, there are two log lines having done a single file update.

```
limitador_1  | [2022-07-13T15:01:43Z INFO  limitador_server] Reloaded limit file
limitador_1  | [2022-07-13T15:01:43Z INFO  limitador_server] Reloaded limit file
```

alexsnaps:
I'll have to check on this double thing… I need more details on the kind of events firing on all platform combination… But reloading twice is better than none I guess…

PR #93, opened 2022-07-20
Issue 64 take2

### LINKED ISSUE #64: Duplicate limits with same namespace, conditions and variables are allowed
Following up from:
- https://issues.redhat.com/browse/OHSS-12001
- https://issues.redhat.com/browse/MGDAPI-3926

It was noticed that there was a duplicate entry in Limitator that targeted the same domain in RHOAM
```
$ oc -n redhat-rhoam-observability-operator rsh observability-operator-controller-manager-856bc6f7bc-pqxvp 
sh-4.4$ curl 10.11.40.16:8080/limits/apicast-ratelimit
[{"namespace":"apicast-ratelimit","max_value":6945,"seconds":60,"name":null,"conditions":["generic_key == slowpath"],"variables":["generic_key"]},{"namespace":"apicast-ratelimit","max_value":70,"seconds":60,"name":null,"conditions":["generic_key == slowpath"],"variables":["generic_key"]}]
```

<details>

```json
[
  {
    "namespace": "apicast-ratelimit",
    "max_value": 6945,
    "seconds": 60,
    "name": null,
    "conditions": [
      "generic_key == slowpath"
    ],
    "variables": [
      "generic_key"
    ]
  },
  {
    "namespace": "apicast-ratelimit",
    "max_value": 70,
    "seconds": 60,
    "name": null,
    "conditions": [
      "generic_key == slowpath"
    ],
    "variables": [
      "generic_key"
    ]
  }
]
```

</details>

RHOAM limits are loaded into limitador by mounting a config map of the limits yaml and setting the `LIMITS_FILE`  env var to this file path [1]. As a customer changes their rate limit quota, this config map is updated and causes a re-scale of limitador pods to pick up the new values.

Typically the old entry is updated / removed in order to use the new value, however in this case another entry was added. This has caused the rate limiting to use the old value rather than the newest value.

[1] https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/marin3r/rateLimitService.go#L173-L202

alexsnaps:
Closes #64 for real?

### DIFF
--- a/limitador/src/limit.rs
+++ b/limitador/src/limit.rs
@@ -26,8 +26,10 @@ impl From<String> for Namespace {
 #[derive(Eq, Debug, Clone, Serialize, Deserialize)]
 pub struct Limit {
     namespace: Namespace,
+    // #[serde(skip_serializing)]
     max_value: i64,
     seconds: u64,
+    // #[serde(skip_serializing)]
     name: Option<String>,
 
     // Need to sort to generate the same object when using the JSON as a key or
--- a/limitador/src/limit.rs
+++ b/limitador/src/limit.rs
@@ -26,10 +26,10 @@ impl From<String> for Namespace {
 #[derive(Eq, Debug, Clone, Serialize, Deserialize)]
 pub struct Limit {
     namespace: Namespace,
-    // #[serde(skip_serializing)]
+    #[serde(skip_serializing)]
     max_value: i64,
     seconds: u64,
-    // #[serde(skip_serializing)]
+    #[serde(skip_serializing)]
     name: Option<String>,
 
     // Need to sort to generate the same object when using the JSON as a key or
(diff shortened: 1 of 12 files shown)


limitador/src/limit.rs
```diff
@@ -26,8 +26,10 @@ impl From<String> for Namespace {
 #[derive(Eq, Debug, Clone, Serialize, Deserialize)]
 pub struct Limit {
     namespace: Namespace,
+    #[serde(skip_serializing)]
```
alexsnaps, line 29: Not sure why I changed that in the first PR, but this still needs to _read_ these values when reading the `LIMITS_FILE` obviously… otherwise all limits are `0` 🤦‍♂️

eguzki (approved):
Just heads up about a **breaking change** here.

Using your own words, the key used to lookup a counter has changed, meaning that upgrading you'd loose your counters as of today.

PR #101, opened 2022-08-02
Full version

alexsnaps:
This adds a more explicit version string (version, as per the `Cargo.toml`, the profile used to build if not `release` and the git sha), to the `--help` & `--version` but also logs it to `stdout` on startup as well as to the logs (`INFO`). The idea is that we have good traceability for when "something goes wrong". 

### Example
<details>

```
$ ./limitador-server ./limitador-server/examples/limits.yaml
Limitador Server debug build v0.5.1 (b3509e30)
```

#### At `INFO` log level, you get the version as well, and the dump of the actual config that'll be used 

```
$ ./limitador-server -vv ./limitador-server/examples/limits.yaml
Limitador Server debug build v0.5.1 (b3509e30)
[2022-08-02T14:01:06Z INFO  limitador_server] Version debug build v0.5.1 (b3509e30)
[2022-08-02T14:01:06Z INFO  limitador_server] Using config: Configuration { limits_file: "./limitador-server/examples/limits.yaml", storage: InMemory, rls_host: "0.0.0.0", rls_port: 8081, http_host: "0.0.0.0", http_port: 8080, limit_name_in_labels: false, log_level: Some(Info) }
```

#### if the directory tree isn't "clean", i.e. has uncommited changes

The `-dirty` suffix is added to the sha
```
$ ./limitador-server ./limitador-server/examples/limits.yaml
Limitador Server debug build v0.5.1 (2a57743d-dirty)
```

#### Logs the profile target

If it's not `release`:

```
$  ./target/debug/limitador-server --version
Limitador Server v0.5.1 (aa83d91e) debug build
```

#### finally, if `git` can't be executed, or this isn't build of a fork

Instead of the sha, you get `unknown`

```
$ ./limitador-server ./limitador-server/examples/limits.yaml
Limitador Server debug build v0.5.1 (unknown)
```
</details>

### DIFF
--- a/limitador-server/build.rs
+++ b/limitador-server/build.rs
@@ -1,34 +1,13 @@
+use std::error::Error;
 use std::process::Command;
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let git_hash = Command::new("git")
-        .args(&["rev-parse", "HEAD"])
-        .output()
-        .ok()
-        .filter(|output| output.status.success())
-        .and_then(|x| String::from_utf8(x.stdout).ok())
-        .map(|hash| hash[..8].to_owned());
+fn main() -> Result<(), Box<dyn Error>> {
+    set_git_hash("LIMITADOR_GIT_HASH");
+    set_profile("LIMITADOR_PROFILE");
+    generate_protobuf()
+}
 
-    let dirty = Command::new("git")
-        .args(&["diff", "--stat"])
-        .output()
-        .ok()
-        .filter(|output| output.status.success())
-        .map(|output| !matches!(output.stdout.len(), 0));
-
-    match git_hash {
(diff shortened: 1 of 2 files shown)


limitador-server/build.rs
```diff
@@ -1,4 +1,37 @@
...
+        .args(&["diff", "--stat"])
+        .output()
+        .ok()
+        .filter(|output| output.status.success())
+        .map(|output| !matches!(output.stdout.len(), 0));
+
+    if Some(true) == dirty && git_hash.is_some() {
+        println!(
+            "cargo:rustc-env=LIMITADOR_GIT_HASH={}-dirty",
+            git_hash.unwrap_or_else(|| "unknown".to_owned())
```
alexsnaps, line 22: May be I should just call `git_hash.unwrap()` here, as there is no way this can be dirty from an "unknown" git tree...

eguzki (approved):

PR #103, opened 2022-08-12
Clippy

alexsnaps:


alexsnaps:
Bypassing review, this is "just" to fix the broken clippy from the new rust release

PR #104, opened 2022-08-15
Clippy

alexsnaps:
So we've had a [few issues](https://github.com/Kuadrant/limitador/pull/103) over the weekend because of the `clippy 0.1.63 (4b91a6ea 2022-08-08)` release.
Most notably, to me at least, the `duplicate_mod` in [`limitador-server/src/envoy_rls/envoy_types.rs`](https://github.com/Kuadrant/limitador/blob/main/limitador-server/src/envoy_rls/envoy_types.rs).

[This fix](https://github.com/Kuadrant/limitador/pull/104/commits/ee97406d0f3a59e1f3e003f8fb87e0be01c50b13) aims at solving this issue. Instead of generating all the sources in our source tree, it now is generated to `target/debug/build/<build>/out` and includes directly into `envoy_types` using the [`tonic::include_proto!`](https://docs.rs/tonic/latest/tonic/macro.include_proto.html) macro. 

 - f8c5bf9 removes the `allow` for the duplicate mod lint
 - b719ef3 has more clippy fixes in `bench.rs` that I had missed.


eguzki (approved):
Autogenerated source code not managed by git :clap:

PR #106, opened 2022-08-22
Dev versioning?

alexsnaps:
This changes the version of the crate to `0.3.0` and the binary's to `1.0.0`… I've added a `-dev` suffix. Trying to reflect what gets build off `main` or other branches, but mostly _not_ at 'tags' for actual releases… so to avoid ever having e.g. a `1.0.0` before and it released, and proposing to increment right after release re-adding the suffix. Not attached to `-dev` tho… just anything that differentiates it from an actual release. 
@guicassolato any point of view? It'd be great to have this across all components, wdyt?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1137,7 +1137,7 @@ checksum = "04c3b4822ccebfa39c02fc03d1534441b22ead323fa0f48bb7ddd8e6ba076a40"
 
 [[package]]
 name = "limitador"
-version = "0.2.0"
+version = "0.3.0-dev"
 dependencies = [
  "async-trait",
  "cfg-if",
@@ -1161,7 +1161,7 @@ dependencies = [
 
 [[package]]
 name = "limitador-server"
-version = "0.5.1"
+version = "1.0.0-dev"
 dependencies = [
  "actix-rt",
  "actix-web",
--- a/limitador-server/Cargo.toml
+++ b/limitador-server/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "limitador-server"
-version = "0.5.1"
+version = "1.0.0-dev"
 authors = ["Alex Snaps <asnaps@redhat.com>", "Eguzki Astiz Lezaun <eguzki@redhat.com>", "David Ortiz <z.david.ortiz@gmail.com>"]
 license = "Apache-2.0"
 keywords = ["rate-limiting", "rate", "limiter", "envoy", "rls"]
--- a/limitador/Cargo.toml
+++ b/limitador/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "limitador"
-version = "0.2.0"
+version = "0.3.0-dev"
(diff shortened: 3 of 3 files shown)


maleck13:
would it make sense to use a something like -rc0 (release candidate 0 )

alexsnaps:
I'm cool with `-rc0` but if we do release candidates, it becomes weird: You are at `-rc0` (i.e. work in progress towards whatever version it suffixes, e.g. `1.0`) then you release `-rc1`, dev branch goes back to `-rc0` so that you have `-rc0 < -rc1 < -rc0`… which is weird (and not how versioning order things… i.e. lexical order, hence `-alpha < -beta`

But with that rational in mind `-dev < -rc` anyways… 🤔

alexsnaps:
@maleck13 I also realize I may have badly conveyed my intent here: `-SNAPSHOT` [people use in Maven](https://maven.apache.org/guides/getting-started/index.html#What_is_a_SNAPSHOT_version)

alexsnaps:
Also, for completeness here, I tried changing the `build.rs` to use ` git tag --points-at HEAD` and figure out some magic about what version it actually builds… but somehow I couldn't get it to work reliably and also thought it wouldn't address it holistically… at the limitador level, if we'd push the output to a file and use that for versioning throughout;… but what about other components? So decided to open the floor for discussion with this change first.

maleck13:
I follow now. So we would have v0.3.0-dev. This would be a build of all the components injecting the version in to the binary and also building and publishing a new image to quay. Then when we felt we were close to a release we would switch to -rc0 etc?
Ideally then this new dev version would also be tested in an environment with a set of very stable (really really stable even if that means we only have 1 or 2 initially) tests that should never fail and are red flag that something happened. I think we should look to trigger these builds automatically on a nightly basis (as a goal).
Once we go through doing it once manually, we can then look to automate it via GH actions or some other mechanism (as long as its not from someones machine).

eguzki (approved):

alexsnaps:
>Then when we felt we were close to a release we would switch to -rc0 etc?

Yes, we can do releases with RCs or whatever however we see fit (including a `-nightly-<date>` or something that's automated… if we feel we want to distribute that), I just would like some marker that means "that's a thing that's built off this branch (could be `main` if e.g. `v1.0.0` is where that happens) but that's _NOT_ the release… 

>So installing the operator at v0.3.0-dev would install the latest -dev releases?

As wrt to dependencies across components… well, that's a bit a mess with `-dev` (or whichever suffix we go for), as it is per definition "unstable", which would mean that you could perfectly use `n-dev` of one component that would end up being incompatible with `n-dev` of one if its dependencies. This could happen for multiple reasons:

 - There is no atomic publishing of these components
 - There is no atomic installation of these neither (e.g. you run `n-dev` of an operator than only "later" pulls in an image of e.g. limitador… that could be weeks later and as such it'd pull the last `n-dev` published then…)

Issue #94, opened 2022-07-22
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 1', limitador/src/limit.rs:131:29


andwang130:
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 1', limitador/src/limit.rs:131:29

alexsnaps:
This is due to a `condition` of a `Limit` that isn't properly formatted. Currently the format is pretty strict: `variable == value` where even the space matter. I have some work on-going that catches those when the limit is loaded, unlike now where it fails _when the limit is evaluated_. I'm also working on making the parsing more robust... 
This should be addressed by the next release

PR #109, opened 2022-08-25
Scanner based parsing of Limit conditions

### LINKED ISSUE #102: Proper expression language (EL) for conditions
Currently `conditions` for a `limit` have the current format:

```yaml
  conditions:
    - "req.method == GET"
```

Where left-hand side (lhs) _has to be_ the `variable` to look up and the right-hand side (rhs) the value it'll be tested against. Currently we _only_ support the equality operator  (`==`) and further more, _only_ the format where it has a white space before _and_ after the operator is [actually supported](https://github.com/Kuadrant/limitador/blob/main/limitador/src/limit.rs#L124-L131) (where more spaces will probably _not_ have [the result one would expect](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1a874510d22ec6c0d27294e5d303d56f)).

I started some work on parsing these condition using a proper scanner which turns the input `&str` (each condition) into tokens (identifier, operator & ... a `String` literal). But right now it still has to fallback to position to know which is the identifier (lhs) and which is the value (rhs). I think that's less optimal and possibly creates really surprising things to read, as e.g. `get == GET` would be a valid condition.

I'm proposing to have `String` literals just be… string literals as one would know them from different programing languages or... yaml for this matter (i.e. surrounded by quotation (`"` or `'`) marks). That would achieve a few things: 

 - White spaces don't matter anymore (or the lack thereof, see #94)
 - Positioning doesn't matter, `method == "GET"` and `"GET" == method` are equivalent, as one would expect as `==` is commutative.
 - Reading the condition is straightforward and less error prone

What would that EL look like tho? 
 - Should we use double (`"`) or/and single (`'`) quotes? 
 - Based on [my reading](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action-generickey), there can only ever be a `String` value in there, so we don't need to support `Number` literals, right? 
 - Use backslah (`\`) for escaping? Using both single & double quotes?

The downside is that would be a breaking change. Existing `limits.yaml` file would be "wrong". I already wanted to eagerly parse the conditions when parsing the `limits.yaml` file (unlike the lazy way of today). So we could give a good error message to users when the scanner would have a `[TokenType::Identifier, TokenType::Operator, TokenType::Identifier]` when it'd expect one (and probably the last one) to be a `TokenType::Literal` instead. 

## tl;dr

Have condition be in the format of:

```yaml
  conditions:
    - "req.method == 'GET'"
```

### todo

 - [ ] Validate conditions _when parsing the limit file_ see #100
 - [ ] Back `TryFrom<&str> for Condition` by a proper `Scanner` impl.

alexsnaps:
Still requires some work to happen in a follow up PR

But this does add support for [this proposal](https://github.com/Kuadrant/limitador/issues/102#issuecomment-1225814778)
 - [x] But this backs the `TryFrom<&str> for Condition` by a `conditions::Scanner` that tokenizes the `String` representation of conditions from the `Limit`
 - [x] Need to support `Number` literals (probably `i64`?), if only to coerce them into `String` when parsing in "backwards compatibility mode", don't want to have `Literal::Identifier` supported that starts with a numeric value…
 - [x] Have more robust error reporting… with scanned `Vec<Token>` available, so we can...
 -  ... ~move the backwards compatibility cases out of the "happy path" into the `Err` path and handle it accordingly~ (see comment below)

Opinionated fix to #94 and closes #102


limitador/src/limit.rs
```diff
@@ -166,13 +330,283 @@ impl PartialEq for Limit {
...
+                literal: Some(Literal::Identifier(
+                    self.input[start - 1..self.pos].iter().collect(),
+                )),
+                pos: start,
+            })
+        }
+
+        fn valid_id_char(&mut self) -> bool {
+            let char = self.input[self.pos];
+            char.is_alphanumeric() || char == '.' || char == '_'
```
alexsnaps, line 431: The dot (`.`) being part of this here makes me sad… I'd expect `foo.bar` to not simply be an identifier…

alexsnaps:
Err… this is unfortunate, looks like we'll need to keep the backwards compatibility path _within_ the happy path:
While serde will give us nice errors:

```
$ ./target/debug/limitador-server limitador-server/examples/limits.yaml
Limitador Server v0.5.1 (5733367b-dirty) debug build
Failed to load limit file: Invalid limit file: Couldn't parse: .[0].conditions: Invalid condition: "req.method == GET"
 └ SyntaxError: Unexpected token `Identifier: GET` at offset 15 at line 7 column 5
```

When using the "deprecated" format for a condition, there is no way to get to the underlying error from the `result::Result<T, Error>` returned by Serde when deserializing a `Limit`. So I can't get to the `ConditionParsingError` that creates this nice error message, as it's been lost (only the `ErrorCode::Message(Box<str>)` is left which contains the _whole error message_). 

I'll look around if there is a way to have some callbacks/error hook registered and do something "elsewhere". But in case this doesn't work, we can rely on the two branches for the fallback into the match statement to support the "deprecated mode" as earlier in this PR's work. It would mean that the lib itself tho would print out to `stderr` with a link to the docs… _for each and every single condition_ in the old format… possibly flooding the logs…

Another option is to bail out on the first error and find some (probably ugly hack involving static) to enable the "lenient mode for the deprecated syntax" from the command line instead… 🤔 if someone has an another idea… I'd love to hear it!

alexsnaps:
In [limits.rs](https://github.com/Kuadrant/limitador/pull/109/files#diff-9d5e22caa49ebe6e41e34bbab7c6a4ede412b1ea93a5b4fac9e12f82ead1bdea) is where it's at 🤦 
Most of the other changes are updating the conditions to use the new syntax…

alexsnaps:
I think I have an idea… "Just" need to implement the `Deserializer` for `Limit` myself … that should do it! I'll give this a try

alexsnaps:
>I think I have an idea… "Just" need to implement the Deserializer for Limit myself … that should do it! I'll give this a try

So while this would "somewhat" work, it still doesn't let me handle the "deprecated case" at the binary `limitador-server` layer. I always end up needing to make the `limitador` lib crate aware of it… 

I will do this a follow up PR tho to make this more digestible to reviewers

limitador/src/limit.rs
```diff
@@ -34,12 +38,224 @@ pub struct Limit {
...
+    error: SyntaxError,
+    pub tokens: Vec<Token>,
+    condition: String,
+}
+
+impl Display for ConditionParsingError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Invalid condition: \"{}\"\n └ {}",
```
alexsnaps, line 66: Given how these are reported by serde:
```
Failed to load limit file: Invalid limit file: Couldn't parse: .[0].conditions: Invalid condition: "req.method == GET"
 └ SyntaxError: Unexpected token `Identifier: GET` at offset 15 at line 7 column 5
```

I'm thinking of suffixing it with ` for condition` so it'd read:
`Unexpected token `Identifier: GET` at offset 15 for condition at line 7 column 5` (as the `line` & `column` is where the condition field is, and `offset` is where the `SyntaxError` within that condition was found)

alexsnaps: Changed in [5f8a405](https://github.com/Kuadrant/limitador/pull/109/commits/5f8a405e83a52e1172d5ecd42d3115b106c48d32) to `Failed to load limit file: Invalid limit file: Couldn't parse: .[0].conditions: SyntaxError: Unexpected token `Identifier: GET` at offset 15 of condition "req.method == GET" at line 7 column 5`

limitador/src/limit.rs
eguzki: Actually, it can be any utf-8 encoded string. Same for the `operand`. This code only applies to the `identifier` which is related to the descriptor entry key.

eguzki (commented):
First of all, I think this is a very good piece of work, well tested. Maybe some doc missing to be very explicit about limits syntax?

I have to comments regarding the PR at high level. Code wise, I could not find anything relevant to request a change, it is brilliant (at least for me).

### scanner

I wonder what is the benefit of implementing a full scanner instead of using a regexp. The limits parsing is not in the critical path of lmitador, i.e. not needed to respond to envoy for the rate limit request. The parsing happens "from time to time", so performance does not seem to be a good reason, is it?

I have been "playing" with a regexp that should provide the same capabilities as the scanner, in terms of parsing. I am aware that a custom scanner allows better error reporting. 

```rust
use regex::Regex;

fn main() {
    let re = Regex::new(r"([\w_\.]+)\s*==\s*'(.+)'").unwrap();

    let conditions_str = [
        "foo=='bar '",
        "foo == 'bar '",
        "  foo == 'bar ' ",
        "  foo   ==   'bar ' ",
        "var == 42",
        " 変数 == '  💖 '",
        "foo == 'ba",
        r#""x != 5 && x > 12""#,
    ];

    for i in 0..conditions_str.len() {
        match re.captures(conditions_str[i]) {
            Some(caps) => println!(
                "{}: identifier: {}, predicate: EQUAL, operand: {}",
                conditions_str[i],
                caps.get(1).unwrap().as_str(),
                caps.get(2).unwrap().as_str(),
            ),
            None => println!("{}: does not parse", conditions_str[i]),
        };
    }
}
```
And running the snippet above, it gives:

```foo=='bar ': identifier: foo, predicate: EQUAL, operand: bar 
foo == 'bar ': identifier: foo, predicate: EQUAL, operand: bar 
  foo == 'bar ' : identifier: foo, predicate: EQUAL, operand: bar 
  foo   ==   'bar ' : identifier: foo, predicate: EQUAL, operand: bar 
var == 42: does not parse
 変数 == '  💖 ': identifier: 変数, predicate: EQUAL, operand:   💖 
foo == 'ba: does not parse
"x != 5 && x > 12": does not parse

```

### the operand is typed

I think that we should handle all `identifiers` (matched with descriptor entry keys) and `operands` (matched with descriptor entry values) as `strings`. First, we do not to break backward compat. Second, and more important IMO, is that the underlying [envoy rate limit protocol](https://github.com/envoyproxy/envoy/blob/7ef98ec1fc3155b83adeaa5adb83e221c4008a59/api/envoy/extensions/common/ratelimit/v3/ratelimit.proto#L81) specifies them as strings. So, IMO, there is no point in requiring quotes (like in yaml) to make string type explicitly. A similar use case are the HTTP headers. The protocol says both header name and the value have to be strings. So when some service process HTTP headers, does not need to know the type. Limitador limits would need numeric support if the underlying envoy rate limit protocol would add support for it, which I do not think it will happen any time soon. Another possibility is that limitador supports other rate limiting protocols besides envoy's one, which also I do not think it will happen any time soon.

alexsnaps:
### Scanner

>The parsing happens "from time to time", so performance does not seem to be a good reason, is it?

No indeed, performance was never the driver. There are I think multiple reasons to _not use_ regular expressions for such a job, [like these here](https://softwareengineering.stackexchange.com/questions/113237/when-you-should-not-use-regular-expressions/113243#113243), but the ones that were the main ones for me here are:

 - [x] Proper error reporting: I wanted something than "this doesn't parse".
 - [x] Extensibility: knowing we wanted to have more operators than just `==`… _and_ operand _types_ possibly (like regex literal for instance

I thought that the "overhead" of writing a scanner which output can then later be fed into a parser and optimizer (mostly to figure out non-sensible configs) was probably worth it. Especially as evolving syntax when you started with regular expressions can become _really_ hard, to [even impossible](https://github.com/Shopify/liquid/issues/138) and I know my regex foo isn't that good that I can be sure it's free of unintended consequences. Regular expression is one of there example where [easy != simple](https://www.youtube.com/watch?v=SxdOUGdseq4). They make it look easy, but the result tends to be very complicated. Tho, _it might not be the case here_ or _yet_. The scanner approach was a way to manage this early on.

### the operand is typed

Right now _everything_ is a String and treated as such. I totally want to keep it that way. What I think you are saying is that anything left and right from the operator goes… Which _could_ be correct. But what are the "value" of "spaces"? What does ` foo ==  bar ` mean? Is it variable ` foo` needs to be equal to ` bar `? So that an operator _has_ to be prefixed and suffixed by a _single_ white space character (` `)? Do we trim? If someone _needs_ to express the white space as being part of a "token", how would that work? I left escaping out of my scanner on purpose for now for these cases if/when they arise. Finally are you implying that we should keep ordering? i.e. left hand side (lhs) is the identifier and right hand side (rhs) is the value _always_? While I am not against it per se, I don't see what forces us to do that… I think it is at least surprising as the only operator(s) (`==` and soon `!=`) are to be symmetric in their very nature. Why would we break that? 

## So what do we do?

 - [ ] If we want to support `any utf-8 encoded string` as you're asking, which _might_ be the correct thing to do, then we need demarcation for identifiers as well as values (e.g. `identifier====value` would be a legal thing to test for… but what does it mean? `identifier==` expected to have value `value` or `identifier=` is `=value` or… you get the idea).
 - [ ] I'd prefer to keep symmetric operator as such, so that position should matter… wdyt?
 - [ ] I'd rather _not_ use regex for what I expressed above… even more so given the above now, but do you feel this scanner approach is actually too complex? But maybe it doesn't matter as
 - [ ] ... expressing conditions as string was maybe the mistake all along:
     - We could have `$"some identifier =" == '=some "value   " '` as an alternative syntax; or 
     - just get rid of syntax altogether:
     
```yaml
condition:
  identifier: "some identifier ="
  eq: '=some "value   " '
```

eguzki (approved):

guicassolato:
PR looks to me.

I do prefer being explicit about the value with the quotes, mainly due to the issue of the white spaces described by @alexsnaps, but also because it feels right. Of course, now you also have the issue of escaping quotes, but still. You cannot have the same syntax for a symbol (e.g. variable name) and for a constant... Unless you work with "tagged tokens" as in structured expressions where the user clearly is requested to provide a _variable name_, an _operator_ and a _constant string value_.

Now a comment about regexes and parsers, for my own sanity here...

A good reason not to use regex is because you have better tools to perform the job (paraphrasing from the reference linked by @alexsnaps). Well, this is true for pretty much anything in life and it also applies here.

Thing is, when we say "regex" to a software engineer and to a computer scientist, they may understand slightly different things, but enough to create a whole confusion. In fact, in the same thread linked by @alexsnaps, the [next answer](https://softwareengineering.stackexchange.com/a/113248) after the one linked is a perfect example of such misunderstanding about regexes.

In the particular case of conditions in Limitador, they _do_ belong to the class of [regular languages](https://en.wikipedia.org/wiki/Regular_language) AFAIK and _could_ be tokenised using regular expressions... buuuuut (no so fast):
1. as long as we can find an implementation of classical regular expressions in Rust instead of the typical regular expression on steroids provided by modern programming languages nowadays, or at least can ensure whoever touches this part of the code never unlocks the power of this flavoured regex implementation and continues forever limiting the expression to the class of regular languages;
2. only if we're wiling to give up on the nice enhancements enumerated by @alexsnaps, like better error messages;
3. Limitador conditions ∈ regular languages is true _today_. Tomorrow, we may want (and I believe we will want!) to break into the class of context free languages for Limitador conditions. For example, balancing parenthesis/braces (just like nested tags) no longer fits into the class of regular languages. If we want to support string templates...

BTW, the association between regexes and performance comes from the scientific definition that equals regular expressions and regular languages. Once we open the door for the "flavoured regex" implementation, linear complexity is no longer guaranteed.

Final note from this who taught compiler theory for a few food years in the university - "scanner" and "parser" are not the same thing. I'm not very knowledgeable in Rust, but it looks like we're implementing a scanner (a simpler task) employing a technique normally used in parsers (more demanding), instead of following the textbook. Nothing wrong with that IMO, for the reasons already mentioned.

guicassolato:
@alexsnaps 

> How would we then support `OR`?

I've been seeing 'and' vs 'or' semantics being represented in YAML/JSON containing lists of rules with a qualified type, and nesting by allowing the qualified type to be referred recursively.

E.g. [Envoy matcher predicates](https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/matcher.proto#envoy-v3-api-msg-xds-type-matcher-v3-matcher-matcherlist-predicate) defines something in the lines of:

```
<predicate> → <single_predicate> | <or_matcher> | <and_matcher> | <not_matcher>
<single_predicate> → the actual expression with operand, operator and value
<or_matcher> → <predicate list>
<and_matcher> → <predicate list>
<not_matcher> → <predicate>
<predicate list> → <predicate> <predicate list> | ε
```

alexsnaps:
@guicassolato 
Tots… the "issue" being that we sort of went down a road already: 

```yaml
  conditions:
    - "req.method == POST"
    - "other == 2"
```

Where, as you can probably guess, this means : `req.method == POST && other == 2` in today's "parlance".

alexsnaps:
Merged #110 on this one, so that nothing should break™ on `main` or images pushed when using the deprecated format

PR #110, opened 2022-08-29
Re-enabled lenient mode

alexsnaps:
This re-adds the support for the "deprecated"/old syntax for conditions: 
 - [x] as a feature in the crate
 - [x] that the server enables, and checks whether some condition did use the deprecated syntax while parsing the config file, and possibly warns the user (despite being logged at `ERROR` level, as this is the default level): 

```
[2022-08-29T20:05:17Z ERROR limitador_server] You are using deprecated syntax for your condition! See guide https://kudrant.io/migration/limitador/constraints
```

The address is: 
 - [ ] open to suggestion, wrt to its parts as well
 - [ ] missing the content that explains what's expected… 

Should we add a "non-lenient" parser option, so the user can validate their newer files? Or make a "only validate limit file" option, that will not start the server but only parse the `limits file`… so that can stay even after this lenient mode gets deprecated, I think @maleck13 proposed something like that actually…

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -211,7 +211,7 @@ impl Limiter {
                             Self::Async(limiter) => limiter.configure_with(limits).await?,
                         }
                         if limitador::limit::check_deprecated_syntax_usages_and_reset() {
-                            error!("You are using deprecated syntax for your condition! See guide https://kudrant.io/migration/limitador/constraints")
+                            error!("You are using deprecated syntax for your conditions! See the migration guide https://kudrant.io/migration/limitador/constraints")
                         }
                         Ok(())
                     }
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -210,6 +210,9 @@ impl Limiter {
                             Self::Blocking(limiter) => limiter.configure_with(limits)?,
                             Self::Async(limiter) => limiter.configure_with(limits).await?,
                         }
+                        if limitador::limit::check_deprecated_syntax_usages_and_reset() {
+                            error!("You are using deprecated syntax for your condition! See guide https://kudrant.io/migration/limitador/constraints")
+                        }
                         Ok(())
                     }
                     Err(e) => Err(LimitadorServerError::ConfigFile(format!(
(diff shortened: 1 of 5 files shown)


limitador-server/src/main.rs
```diff
@@ -210,6 +210,9 @@ impl Limiter {
                             Self::Blocking(limiter) => limiter.configure_with(limits)?,
                             Self::Async(limiter) => limiter.configure_with(limits).await?,
                         }
+                        if limitador::limit::check_deprecated_syntax_usages_and_reset() {
+                            error!("You are using deprecated syntax for your condition! See guide https://kudrant.io/migration/limitador/constraints")
```
alexsnaps, line 214: >See guide

See how to migrate with this guide?

limitador-server/src/main.rs
```diff
@@ -210,6 +210,9 @@ impl Limiter {
                             Self::Blocking(limiter) => limiter.configure_with(limits)?,
                             Self::Async(limiter) => limiter.configure_with(limits).await?,
                         }
+                        if limitador::limit::check_deprecated_syntax_usages_and_reset() {
```
eguzki, line 213: if the feature is not enabled, this would not compile

eguzki (changes requested):

limitador-server/src/main.rs
alexsnaps: Sure... but the feature is always enabled. The way I've done this is the crate has the feature, and this enables it _always_ as a consumer of the crate.
Are you asking I do make the feature available at the binary's layer too (like I did for infinispan)?

eguzki: ok I see.

The feature belongs to the `limitador` crate. And the limitador crate compiles regardless of being activated or not. However, limitador-server package needs it activated or it would not compile.

LGTM

eguzki (approved):
LGTM

Regarding the "ugly" static switch, It is ok as it is, but I would have added one error per condition, possibly flooding the log. IMO it is a fair toll to pay when using deprecated syntax.

Eventually, in the next release or after, when the deprecated syntax is dropped (or entirely changed by a new struct), this "hack" can be removed

alexsnaps:
>Regarding the "ugly" static switch, It is ok as it is, but I would have added one error per condition, possibly flooding the log. IMO it is a fair toll to pay when using deprecated syntax.

So, to stderr from the crate itself directly? Or have a static `Vec<String>` with all the deprecated style conditions, so the logger can log them all (rather than the current generic message)?

>Eventually, in the next release or after, when the deprecated syntax is dropped (or entirely changed by a new struct), this "hack" can be removed

Yes, this goes away in a minor release to come, totally!

PR #112, opened 2022-08-30
Updated dependencies

alexsnaps:
- [x] froze [notify crate](https://docs.rs/notify/5.0.0-pre.15/notify/)
- [x] updated all deps requiring no code changes
- [x] update homepage to kuadrant.io
- [x] updated to latest rustc toolchain
- [x] removed `libssl3` from container image, as this was introduced by `reqwest`/`infinispan`
- [ ] updated [`authors`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field)… but it isn't complete… confused?
  - Do we want active authors or _all_ authors… I'd argue for the latter!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,6 +22,9 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+      - uses: abelfodil/protoc-action@v1
+        with:
+          protoc-version: '3.19.4'
       - uses: actions-rs/cargo@v1
         with:
           command: check
@@ -47,6 +50,9 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+      - uses: abelfodil/protoc-action@v1
+        with:
+          protoc-version: '3.19.4'
       - uses: actions-rs/cargo@v1
         with:
           command: test
@@ -79,6 +85,9 @@ jobs:
           toolchain: stable
           override: true
       - run: rustup component add clippy
+      - uses: abelfodil/protoc-action@v1
+        with:
+          protoc-version: '3.19.4'
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
@@ -97,6 +106,9 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+      - uses: abelfodil/protoc-action@v1
+        with:
+          protoc-version: '3.19.4'
       - uses: actions-rs/cargo@v1
         with:
           command: bench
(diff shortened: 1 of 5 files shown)


eguzki (commented):
LGTM so far

limitador/Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 name = "limitador"
 version = "0.2.0"
-authors = ["David Ortiz <z.david.ortiz@gmail.com>"]
+authors = ["David Ortiz <z.david.ortiz@gmail.com>", "Eguzki Astiz Lezaun <eguzki@redhat.com>", "Alex Snaps <asnaps@redhat.com>"]
 license = "Apache-2.0"
 keywords = ["rate-limiting", "rate", "limiter"]
 categories = ["web-programming"]
 description = "Rate limiting library"
-homepage = "https://github.com/kuadrant/limitador"
+homepage = "https://kuadrant.io"
```
eguzki, line 9: not sure about this, since this crate is publish in crates.io

Aligned with the [previous](https://github.com/Kuadrant/limitador/pull/112/files#r958547115) comment, I would leave it as it was, but it is ok as you propose.

limitador-server/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "Apache-2.0"
 keywords = ["rate-limiting", "rate", "limiter", "envoy", "rls"]
 categories = ["web-programming"]
 description = "Rate limiting service that integrates with Envoy's RLS protocol"
-homepage = "https://github.com/kuadrant/limitador"
+homepage = "https://kuadrant.io"
```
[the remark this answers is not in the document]
alexsnaps: Which ever… we can also _not_ have a homepage (also, to be clear, there is repository that points to github still)

limitador-server/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "Apache-2.0"
...
 repository = "https://github.com/kuadrant/limitador"
 documentation = "https://docs.rs/limitador"
```
alexsnaps, line 11: I'd probably get rid of the `documentation` here, or have it point to [kuadrant.io/docs/limitador-server](https://kuadrant.io/docs/limitador-server/)? (There will be better content there soon™)

limitador-server/Cargo.toml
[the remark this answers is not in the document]
eguzki: :shrug: your call

.github/workflows/rust.yml
```diff
@@ -22,6 +22,9 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+      - uses: abelfodil/protoc-action@v1
```
eguzki, line 25: just for curiosity, why is this new step needed now?

eguzki (approved):

.github/workflows/rust.yml
alexsnaps: https://github.com/hyperium/tonic/issues/1047#issuecomment-1222508191

PR #120, opened 2022-09-01
Resolve limits_file_dir from the canonicalized limit_file

### LINKED ISSUE #119: limits file passed without directory, fails
There is a bug in the way we resolve the directory to watch for changes on the limits file.
If you start limitador-server with a "local to the dir" limits file, the server panics:

```
./target/debug/limitador-server test.yaml
Limitador Server v1.0.0-dev (edddfbce-dirty)  debug build
Error: Error { kind: PathNotFound, paths: [""] }
```

This is because of how we extract the parent directory: `Path::new(&limit_file).parent().unwrap()` where parent is now `""`
We need to resolve the absolute path before doing this…

alexsnaps:
closes #119 

Somehow using the `fs::canonicalize(&limit_file)` to resolve the parent directory fails monitoring the changes in a k8s cluster. I'm not real sure why… I need to understand how k8s maps this `ConfigMap` and remaps it on changes… 
In the meantime, if we didn't resolve a absolute path from the `Path::new(&limit_file).parent()` we then fall back to use the parent dir…

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -269,7 +269,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let limiter = Arc::clone(&rate_limiter);
     let handle = Handle::current();
     // it should not fail because the limits file has already been read
-    let limits_file_dir = Path::new(&limit_file).parent().unwrap();
+    let mut limits_file_dir = Path::new(&limit_file).parent().unwrap();
+    if limits_file_dir.as_os_str().is_empty() {
+        limits_file_dir = Path::new(".");
+    }
     let limits_file_path_cloned = limit_file.to_owned();
     // structure needed to keep state of the last known canonical limits file path
     let mut last_known_canonical_path = fs::canonicalize(&limit_file).unwrap();
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -57,7 +57,7 @@ jobs:
           command: |
             ip=$(kubectl get nodes -lkubernetes.io/hostname!=kind-control-plane -ojsonpath='{.items[*].status.addresses[?(@.type=="InternalIP")].address}')
             port=$(kubectl get service limitador -o jsonpath='{.spec.ports[?(@.name=="http")].nodePort}')
-            curl "http://${ip}:${port}/limits/test" -o output-limits.json 2>/dev/null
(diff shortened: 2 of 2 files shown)


limitador-server/src/main.rs
```diff
@@ -265,12 +265,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let limits_file_dir = Path::new(&limit_file).parent().unwrap();
-    let limits_file_path_cloned = limit_file.to_owned();
+
     // structure needed to keep state of the last known canonical limits file path
     let mut last_known_canonical_path = fs::canonicalize(&limit_file).unwrap();
 
+    // This isn't probably the best fix, see https://github.com/Kuadrant/limitador/issues/119
+    let mut parent = Path::new(&limit_file).parent().unwrap();
+    if !parent.is_absolute() {
+        parent = last_known_canonical_path.as_path().parent().unwrap();
```
eguzki, line 275: This is approach will not work in k8s. In k8s the file structure is as follows:

```
/home/limitador/bin $ ls -al /home/limitador/etc                                                     
total 0                                                                                              
drwxrwsrwx    3 root     10007600       100 Jul 20 22:25 .                                           
drwxr-sr-x    1 limitado limitado        17 Jul 20 22:25 ..                                          
drwxr-sr-x    2 root     10007600        47 Jul 20 22:25 ..2022_07_20_22_25_07.358335008             
lrwxrwxrwx    1 root     10007600        31 Jul 20 22:25 ..data -> ..2022_07_20_22_25_07.358335008                    
lrwxrwxrwx    1 root     10007600        28 Jul 20 22:25 limitador-config.yaml -> ..data/limitador-config.yaml
```

Thus, `last_known_canonical_path` would have a value similar (in structure) as `/home/limitador/etc/..2022_07_20_22_25_07.358335008/limitador-config.yaml`. The parent would be `/home/limitador/etc/..2022_07_20_22_25_07.358335008`, and that is not the directory we want to monitor. We want to monitor the `/home/limitador/etc` directory.

eguzki: I mean that in k8s having a relative "local to the current directory" approach would not work. But using the operator, this is never the case, as the operator always uses absolute path (possibly full of symlinks)

alexsnaps:
@eguzki I think that's it, wdyt?

eguzki (approved):

PR #121, opened 2022-09-01
Validate limits

alexsnaps:
This introduces the option `--validate` to the `limitador-server` executable.

It does validate the `LIMITS_FILE` passed as argument against the expected syntax. 
Prints any error to `stderr`
Prints the validated output back to `stdout`

<details>
## When there is an error:

```
 ./target/debug/limitador-server --validate limitador-server/examples/limits.yaml > new.yaml
Invalid file: ConfigFile("Couldn't parse: .[0]: missing field `seconds` at line 3 column 3")
```

## When there is a condition error:

```
./target/debug/limitador-server --validate limitador-server/examples/limits.yaml > new.yaml
Invalid file: ConfigFile("Couldn't parse: .[0].conditions: SyntaxError: Invalid character `~` at offset 12 of condition \"req.method ~= GET\" at line 7 column 5")
```

## When used to port deprecated syntax to the new:

```
./target/debug/limitador-server --validate limitador-server/examples/limits.yaml > new.yaml
Deprecated syntax for conditions corrected!
```

with ` limitador-server/examples/limits.yaml`

```yaml
---
-
  namespace: test_namespace
  max_value: 10
  seconds: 60
  conditions:
    - "req.method == GET"
  variables:
    - user_id
-
  namespace: test_namespace
  max_value: 5
  seconds: 60
  conditions:
    - "req.method == POST"
  variables:
    - user_id
```

and the resulting ported `new.yaml`

```yaml
- namespace: test_namespace
  max_value: 10
  seconds: 60
  name: null
  conditions:
  - req.method == "GET"
  variables:
  - user_id
- namespace: test_namespace
  max_value: 5
  seconds: 60
  name: null
  conditions:
  - req.method == "POST"
  variables:
  - user_id
```

</details>


alexsnaps:
See #122 , but should I add the check of `limit.max_value > 0` as an addition check as well as part of this?

eguzki (changes requested):
LGTM

limitador-server/examples/limits.yaml
```diff
@@ -12,6 +12,6 @@
   max_value: 5
   seconds: 60
   conditions:
-    - "req.method == 'POST'"
+    - "req.method ~= 'POST'"
```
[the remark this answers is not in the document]
alexsnaps: Ooops… Yeah, that was not meant to be committed 🤦

alexsnaps:
@eguzki fixed your catch… and added 894fcba - let me know what you think… This will not allow zero (`0`) and negative values for `Limit.max_value`… Tho I still don't know why this isn't a `u64` to begin with… 🤔 Or did I miss/forget something?

alexsnaps:
@eguzki and I think this addresses all we've discussed in the status meeting

limitador-server/src/main.rs
```diff
@@ -205,16 +205,22 @@ impl Limiter {
...
                 let parsed_limits: Result<Vec<Limit>, _> = serde_yaml::from_reader(f);
                 match parsed_limits {
-                    Ok(limits) => {
-                        match &self {
-                            Self::Blocking(limiter) => limiter.configure_with(limits)?,
-                            Self::Async(limiter) => limiter.configure_with(limits).await?,
-                        }
-                        if limitador::limit::check_deprecated_syntax_usages_and_reset() {
-                            error!("You are using deprecated syntax for your conditions! See the migration guide https://kudrant.io/migration/limitador/constraints")
+                    Ok(limits) => match first_negative(&limits) {
```
eguzki, line 208: missing some test with negative values?

limitador-server/src/main.rs
```diff
@@ -462,6 +477,12 @@ fn create_config() -> (Configuration, String) {
                 .display_order(6)
                 .help("Sets the level of verbosity"),
         )
+        .arg(
+            Arg::with_name("validate")
```
eguzki, line 481: This will be documented as part of the limitador-server v1.0.0 crate, right?

eguzki (changes requested):

limitador-server/src/main.rs
alexsnaps: There is no such crate. But yes, there is task on-going to document _all_ these things. This will also be in the migration guide.

limitador-server/src/main.rs
alexsnaps: I can add a test for that function being called if that's what you mean

limitador-server/src/main.rs
```diff
@@ -552,6 +573,41 @@ fn create_config() -> (Configuration, String) {
...
 
     let limits_file = matches.value_of("LIMITS_FILE").unwrap();
+
+    if matches.is_present("validate") {
+        let error = match std::fs::File::open(limits_file) {
+            Ok(f) => {
+                let parsed_limits: Result<Vec<Limit>, _> = serde_yaml::from_reader(f);
+                match parsed_limits {
+                    Ok(limits) => {
+                        if limitador::limit::check_deprecated_syntax_usages_and_reset() {
```
eguzki, line 583: should `first_negative` be called to check existence of negative maxValues?

alexsnaps: Well it does return the first negative `max_value` as an `Option<usize>`

alexsnaps: `find_first_negative_limit` maybe?

eguzki: my point is, the validation command should also check there is no "negative limits", right?

alexsnaps: Oh! Yes!!!

eguzki (approved):

PR #124, opened 2022-09-06
Trying `Limit.max_value` being unsigned

alexsnaps:
This is by no mean complete or possibly even entirely correct, I brute-forced this in a bit… but I wanted to try figuring out why `Limit.max_value` is _signed_ … and, well, I still don't know. Other than possibly integer arithmetic being "annoying" ?

### DIFF
--- a/limitador-server/src/http_api/request_types.rs
+++ b/limitador-server/src/http_api/request_types.rs
@@ -18,7 +18,7 @@ pub struct CheckAndReportInfo {
 #[derive(Debug, Eq, PartialEq, Serialize, Deserialize, Apiv2Schema)]
 pub struct Limit {
     namespace: String,
-    max_value: i64,
+    max_value: u64,
     seconds: u64,
     name: Option<String>,
     conditions: Vec<String>,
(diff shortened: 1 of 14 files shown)


alexsnaps:
Alright, so Redis stores Integer as signed 64 bits and `INCR` is also on `64 bit signed integer` only. 

>`An error was signalled by the server: Error running script (call to f_32b544ae1f80cd9e6bdc62f5b477af978012c803): @user_script:3: ERR value is not an integer or out of range `

because of [this change](https://github.com/Kuadrant/limitador/pull/124/files#diff-e2563135e37b4334c6801174dd8c537e7df15441f334a99bffefc35ffe1e7447R234)

Having i64::MAX_VALUE as an upper limit tho, doesn't invalidate using a `u64` to store it… I don't know what's best tbh… #121 like approach or… this here… or having our own type?

alexsnaps:
That's the [take away](https://github.com/Kuadrant/limitador/pull/124#issuecomment-1239303629) from this, we'll keep things as is in the crate for now and validate while parsing the file see #121

PR #125, opened 2022-09-08
Fallback to github sha, if it can't be resolved using `git`

alexsnaps:
Before this change, we'd lose the sha this was build from and the version would report `unknown`:

```sh
❯ docker run quay.io/kuadrant/limitador limitador-server --help
Limitador Server v1.0.0-dev (unknown)
The Kuadrant team - github.com/Kuadrant
```

But now `b9d2e48e`

```sh
❯ docker run quay.io/kuadrant/limitador:github_sha limitador-server --help
Limitador Server v1.0.0-dev (b9d2e48e)
The Kuadrant team - github.com/Kuadrant
```

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -32,6 +32,8 @@ jobs:
           tags: ${{ env.IMG_TAGS }}
           dockerfiles: |
             ./Dockerfile
+          build-args: |
+            GITHUB_SHA=${{ github.sha }}
       - name: Push Image
         if: ${{ !env.ACT }}
         id: push-to-quay
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,8 @@ COPY ./Cargo.toml ./Cargo.toml
 COPY limitador/Cargo.toml ./limitador/Cargo.toml
 COPY limitador-server/Cargo.toml ./limitador-server/Cargo.toml
 
+ARG GITHUB_SHA
+ENV GITHUB_SHA=${GITHUB_SHA:-unknown}
 ENV RUSTFLAGS="-C target-feature=-crt-static"
 
 RUN mkdir -p limitador/src limitador-server/src
--- a/limitador-server/build.rs
+++ b/limitador-server/build.rs
@@ -33,7 +33,7 @@ fn set_profile(env: &str) {
 }
 
 fn set_git_hash(env: &str) {
-    let git_hash = Command::new("git")
+    let git_hash = Command::new("/usr/bin/git")
         .args(&["rev-parse", "HEAD"])
         .output()
         .ok()
@@ -55,6 +55,9 @@ fn set_git_hash(env: &str) {
             _ => unreachable!("How can we have a git hash, yet not know if the tree is dirty?"),
         }
     } else {
-        println!("cargo:rustc-env={}=unknown", env);
+        let fallback = option_env!("GITHUB_SHA")
+            .map(|sha| if sha.len() > 8 { &sha[..8] } else { sha })
+            .unwrap_or("NO_SHA");
+        println!("cargo:rustc-env={}={}", env, fallback);
     }
 }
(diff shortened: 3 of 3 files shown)


.github/workflows/build-image.yaml
```diff
@@ -32,6 +32,8 @@ jobs:
           tags: ${{ env.IMG_TAGS }}
           dockerfiles: |
             ./Dockerfile
+          build-args: |
+            GITHUB_SHA=${{ github.sha }}
```
alexsnaps, line 36: 1/ Pass the sha that triggered this build to the docker build image

Dockerfile
```diff
@@ -18,6 +18,8 @@ COPY ./Cargo.toml ./Cargo.toml
 COPY limitador/Cargo.toml ./limitador/Cargo.toml
 COPY limitador-server/Cargo.toml ./limitador-server/Cargo.toml
 
+ARG GITHUB_SHA
+ENV GITHUB_SHA=${GITHUB_SHA:-unknown}
```
alexsnaps, line 22: 2/ get it from the CI

limitador-server/build.rs
```diff
@@ -50,11 +50,14 @@ fn set_git_hash(env: &str) {
...
         match dirty {
-            Some(true) => println!("cargo:rustc-env={}={}-dirty", env, hash),
-            Some(false) => println!("cargo:rustc-env={}={}", env, hash),
+            Some(true) => println!("cargo:rustc-env={}={}-dirty", env, sha),
+            Some(false) => println!("cargo:rustc-env={}={}", env, sha),
             _ => unreachable!("How can we have a git hash, yet not know if the tree is dirty?"),
         }
     } else {
-        println!("cargo:rustc-env={}=unknown", env);
+        let fallback = option_env!("GITHUB_SHA")
```
alexsnaps, line 58: 3/ finally use it

eguzki (approved):
LGTM

The only caveat is that if you build the image locally (`docker build -t tag .`) the sha will not be available (unless you pass the GITHUB_SHA with --build-arg). But fair enough

alexsnaps:
> The only caveat is that if you build the image locally (`docker build -t tag .`) the sha will not be available (unless you pass the GITHUB_SHA with --build-arg). But fair enough

I agree, but I don't see/know how to automate that step without being explicit… I searched, but couldn't find a way :( 
_Very_ open to suggestions tho!

eguzki:
I am ok with the content of the PR. 

One way would be, adding git package in the docker stage where it is being built, [here](https://github.com/Kuadrant/limitador/blob/474bec2ca03ebc0cae8e6a28bc31fda5a845601f/Dockerfile#L10) and allowing `.git` being copied removing `.git` from the `.dockerignore` [here](https://github.com/Kuadrant/limitador/blob/474bec2ca03ebc0cae8e6a28bc31fda5a845601f/.dockerignore#L5). That still needs the env var `GITHUB_SHA` because in the CI pipeline, `git diff --stat` leads to `dirty` build. Then when CI pipeline is run, build.rs should check the `GITHUB_SHA` env var. If it exists, it uses that instead of reading from the git command. 

But I think it is complicating things for a tiny thing of having the sha when building locally.

alexsnaps:
>That still needs the env var `GITHUB_SHA` because in the CI pipeline, `git diff --stat` leads to `dirty` build

Why would it be dirty tho? What's being modified?

eguzki:
> Why would it be dirty tho? What's being modified?

I was expecting not to be, but I run `git diff --stat` just after the checkout and it turned out to be "dirty"

eguzki:
> Why would it be dirty tho? What's being modified?

Because of [this](https://github.com/Kuadrant/limitador/blob/85d9cf4efa4d44655c91b5d94f0997a2c277d12a/Dockerfile#L27)

I have the changes I mentioned earlier  in this branch, so you can have a look if you want: https://github.com/Kuadrant/limitador/compare/add-git-build-stage?expand=1

PR #130, opened 2022-10-11
fix(error): Bubble error up to main when connecting to Redis fails

### LINKED ISSUE #100: Validate complete environment on startup
See #99 & #94 as example of current issues. Ideally limitador validated it'll properly respond to requests fine as it boot straps.

Given #108 , once it has been implemented and #76 as well, this should be pretty good…

alexsnaps:
No magic here… 

Fixes #100

### DIFF
--- a/limitador/src/storage/redis/redis_async.rs
+++ b/limitador/src/storage/redis/redis_async.rs
@@ -149,7 +149,8 @@ impl AsyncRedisStorage {
         let info = ConnectionInfo::from_str(redis_url)?;
         Ok(Self {
             conn_manager: ConnectionManager::new(
-                redis::Client::open(info).expect("Somehow couldn't create Redis client!"),
+                redis::Client::open(info)
+                    .expect("This couldn't fail in the past, yet now it did somehow!"),
             )
             .await?,
         })
--- a/limitador/src/storage/redis/redis_async.rs
+++ b/limitador/src/storage/redis/redis_async.rs
@@ -173,3 +173,25 @@ impl AsyncRedisStorage {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::storage::redis::AsyncRedisStorage;
+    use redis::ErrorKind;
+
+    #[tokio::test]
+    async fn errs_on_bad_url() {
+        let result = AsyncRedisStorage::new("cassandra://127.0.0.1:6379").await;
+        assert!(result.is_err());
+        assert_eq!(result.err().unwrap().kind(), ErrorKind::InvalidClientConfig);
+    }
+
+    #[tokio::test]
+    async fn errs_on_connection_issue() {
+        let result = AsyncRedisStorage::new("redis://127.0.0.1:21").await;
+        assert!(result.is_err());
+        let error = result.err().unwrap();
+        assert_eq!(error.kind(), ErrorKind::IoError);
+        assert!(error.is_connection_refusal())
+    }
+}
--- a/limitador/src/storage/redis/redis_async.rs
+++ b/limitador/src/storage/redis/redis_async.rs
@@ -8,7 +8,7 @@ use crate::storage::keys::*;
 use crate::storage::redis::scripts::SCRIPT_UPDATE_COUNTER;
 use crate::storage::{AsyncCounterStorage, Authorization, StorageErr};
 use async_trait::async_trait;
-use redis::AsyncCommands;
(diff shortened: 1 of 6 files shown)


eguzki (commented):
What about the redis storage sync version? If the connection cannot be established, the error is panicked. That is what we wanted to avoid, right?

limitador/src/storage/redis/redis_async.rs
```diff
@@ -145,14 +145,14 @@ impl AsyncCounterStorage for AsyncRedisStorage {
...
 
 impl AsyncRedisStorage {
-    pub async fn new(redis_url: &str) -> Self {
-        Self {
+    pub async fn new(redis_url: &str) -> Result<Self, RedisError> {
+        let info = ConnectionInfo::from_str(redis_url)?;
+        Ok(Self {
             conn_manager: ConnectionManager::new(
-                redis::Client::open(ConnectionInfo::from_str(redis_url).unwrap()).unwrap(),
+                redis::Client::open(info).expect("Somehow couldn't create Redis client!"),
```
eguzki, line 152: Am I right expecting that the function `ConnectionManager::new` will return error when the connection cannot be established?  Because, if it is `redis::Client::open(info)` it will panic and `AsyncRedisStorage::new` will not return error, but panic

limitador/tests/integration_tests.rs
```diff
@@ -38,12 +38,12 @@ macro_rules! test_with_all_storage_impls {
             #[tokio::test]
             #[serial]
             async fn [<$function _with_async_redis>]() {
-                let storage = AsyncRedisStorage::new("redis://127.0.0.1:6379").await;
+                let storage = AsyncRedisStorage::new("redis://127.0.0.1:6379").await.expect("We need a Redis running locally");
```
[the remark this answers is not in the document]
alexsnaps: We don't but I can add one

limitador/src/storage/redis/redis_async.rs
alexsnaps: So that one is really weird: 

```rust
    /// Connects to a redis server and returns a client.  This does not
    /// actually open a connection yet but it does perform some basic
    /// checks on the URL that might make the operation fail.
    pub fn open<T: IntoConnectionInfo>(params: T) -> RedisResult<Client> {
        Ok(Client {
            connection_info: params.into_connection_info()?,
        })
    }
```

and 

```rust
impl IntoConnectionInfo for ConnectionInfo {
    fn into_connection_info(self) -> RedisResult<ConnectionInfo> {
        Ok(self)
    }
}
```

I think this is a less optimal API into the client, but what we do, create the `ConnectionInfo` prior to `open` (which doesn't open), can't ever fail. I added the `expect` there, in case this API's implementation changes in the future.

alexsnaps:
> What about the redis storage sync version? If the connection cannot be established, the error is panicked. That is what we wanted to avoid, right?

Yeah, I mostly cared about the server binary, which leverages the async version. But yeah, I should totally do the sync version as well… otherwise this is weird.

limitador/src/storage/redis/redis_sync.rs
```diff
@@ -199,8 +205,21 @@ mod test {
...
+        assert!(result.is_err());
+        assert_eq!(result.err().unwrap(), "Redis URL did not parse".to_string())
+    }
+
+    #[test]
+    fn errs_on_connection_issue() {
+        // this panic!s And I really don't see how to bubble the redis error back up:
+        // r2d2 consumes it
+        // RedisError are not publicly constructable
+        // So using String as error type… sad
```
alexsnaps, line 219: See #131

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -191,12 +191,9 @@ impl CachedRedisStorage {
...
         ttl_ratio_cached_counters: u64,
-    ) -> Self {
-        let redis_conn_manager = ConnectionManager::new(
-            redis::Client::open(ConnectionInfo::from_str(redis_url).unwrap()).unwrap(),
-        )
-        .await
-        .unwrap();
+    ) -> Result<Self, RedisError> {
+        let info = ConnectionInfo::from_str(redis_url)?;
+        let redis_conn_manager = ConnectionManager::new(redis::Client::open(info).unwrap()).await?;
```
eguzki, line 196: nitpick: maybe `expect` instead of `unwrap` for consistency?

eguzki (approved):

PR #133, opened 2022-10-21
Crate v0.3.0 rc1

alexsnaps:
This is a first shot at trying to publish to crates.io using github actions

 - This is a manually triggered action, where you provide the version: `0.3.0-rc1` in this case
 - It should check the branch out (prefixed with `crate-v`)
 - Test it
 - If all goes well try a `--dry-run` to push to crates.io


.github/workflows/release.yaml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          ref: crate-v${{ github.event.inputs.version }}
+      - name: Build
+        run: cargo build --verbose --release
+      - name: Check
+        run: cargo fmt --all -- --check && cargo clippy -- -D warnings
+      - name: Run tests
+        run: cargo test --verbose
+      - name: Publish
+        if: ${{ success() }}
```
eguzki, line 27: this is really needed? shouldn't the workflow end if a previous step fails?

eguzki (approved):

.github/workflows/release.yaml
alexsnaps: I wasn't sure tbh

alexsnaps: I think you are right and this can be deleted, I'll do it when I get rid of the `--dry-run`

PR #139, opened 2022-10-21
Add version to package spec used to publish

alexsnaps:
With adding the version to the package being released, if the `Cargo.toml` `version` isn't matching, the publish would fail:

```
cargo publish -p limitador:0.3.0
error: package ID specification `limitador@0.3.0` did not match any packages
Did you mean one of these?

  limitador@0.3.0-rc1

```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,4 +30,4 @@ jobs:
       - name: Run tests
         run: cargo test --verbose
       - name: Publish
-        run: cargo publish --token ${{ secrets.CARGO_REGISTRY_TOKEN }} -p limitador
+        run: cargo publish --token ${{ secrets.CARGO_REGISTRY_TOKEN }} -p limitador:${{ github.event.inputs.version }}


.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,46 @@
+---
+name: Release server container image
```
eguzki, line 2: This is pretty much the same as `.github/workflows/build-image.yaml`. Is there a way to reuse the existing build image action spec adding the `on.workflow_dispatch` to `.github/workflows/build-image.yaml`?

alexsnaps: I don't know how to compose these workflows… do you have any pointers? I'll look around

.github/workflows/release-image.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          ref: server-v${{ github.event.inputs.version }}
```
alexsnaps, line 24: This is a non-neglectable change from the `build-image.yaml`…
The rest is configuration changes, so we'd need to add logic to configure one way or the other depending on the way the script was triggered… Wonder if logic is more desirable than plain declarative workflows… but I'll let you be the judge. At this stage I think it's pretty even one way or the other

.github/workflows/release-image.yaml
alexsnaps: Ok, I'll try a `composite` action, refactoring the common bits in `.github/actions/quay.io-publish.yaml` or something. This might need a few trials tho 🤦

.github/workflows/release-image.yaml
eguzki: I wonder if we can avoid both. Let me try to illustrate how I see it.

The only manual action should be tagging the repo. Then tagging itself should trigger other workflows like docker image build and publish crate to crate.io. So, basically, fot the docker images we already have it in `build-image.yaml`

```yaml
on: 
  push:
    branches: [*]
    tags: [*]
```

The "Release crate" workflow should also be triggered by tagging, not by manually triggered workflow. We may want to define better the triggering to only publish a crate for some kind of tags, like `vX.Y.Z` or whatever. Then allow adding other tags will not publish crates.

alexsnaps:
I think this is superseded by the work that @didierofrivia is going to consolidate the infra

PR #146, opened 2023-01-17
Initial persistent storage using Sled

alexsnaps:
Missing pieces:
 - [x] Wire the `Store` as an option in limitador-server
 - [x] Use proper configuration of backend storage rather than `.default()`

Issues:
 - [x] Fix bench for each `Storage` to use a different fresh directory
 - [x] Review key format and better handle iteration when retrieving counters in bulk, see PR #167 
 - [x] Fix `#[cfg()]`... some functions are useless as of now
 - [x] Change `From<>` for our persistent types into `TryFrom<>`, and deal with the errors? Like dropping the counter?

Perf wise, this isn't too bad:

## in-memory check_rate_limited_and_update/10 namespaces with 50 limits each with 10 conditions and 10 variables
image

## disk persistent check_rate_limited_and_update/10 namespaces with 50 limits each with 10 conditions and 10 variables
image

[with](https://github.com/Kuadrant/limitador/pull/146/files#diff-e2563135e37b4334c6801174dd8c537e7df15441f334a99bffefc35ffe1e7447R109):

```rust
DiskStorage::open(tmp.path(), OptimizeFor::Throughput)
```

<details>
Which (on my MacOS with Redis in a docker… so... ヽ(。_°)ノ meh…) looks like this with Redis…

image

… and… yes… these are ms now… not µs anymore.
</details>

### DIFF
--- a/limitador/src/storage/keys.rs
+++ b/limitador/src/storage/keys.rs
@@ -17,8 +17,8 @@ use crate::limit::Limit;
 
 pub fn key_for_counter(counter: &Counter) -> String {
     format!(
-        "namespace:{{{}}},counter:{}",
-        counter.namespace().as_ref(),
+        "{},counter:{}",
+        prefix_for_namespace(counter.namespace().as_ref()),
         serde_json::to_string(counter).unwrap()
     )
 }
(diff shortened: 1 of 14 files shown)


alexsnaps:
Seeing [this only](https://github.com/spacejam/sled/issues/1417#issuecomment-1264310658) now tho

alexsnaps:
> Seeing [this only](https://github.com/spacejam/sled/issues/1417#issuecomment-1264310658) now tho

Because of that I hid Sled away completely in this [commit](a66bafb77cd877367224c76e9b27a22f940ee3db) and expose [some config bits](bf5694e) in our own dedicated types from `crate::storage::disk`

limitador/src/storage/keys.rs
```diff
@@ -48,10 +49,23 @@ pub fn counter_from_counter_key(key: &str, limit: &Limit) -> Counter {
     counter
 }
 
+pub fn partial_counter_from_counter_key(key: &str, namespace: &str) -> Counter {
+    let offset = ",counter:".len();
+    let start_pos_counter = prefix_for_namespace(namespace).len() + offset;
+
+    let counter: Counter = serde_json::from_str(&key[start_pos_counter..]).unwrap();
```
didierofrivia, line 56: Here we are Ok to panic or it's just not possible?

eguzki:
started looking at this :eyes:

eguzki (approved):
Looking super good! 

Unfortunately I do not have enough background knowledge to do a proper review and I would need face2face review. Specially for the implementation of the `CounterStorage` trait

alexsnaps:
Closed in favor of #178

PR #150, opened 2023-03-09
feat: build a multi-arch image adding arm64 support

chirino:


alexsnaps (approved):
Let's try this! Thanks for the PR

PR #153, opened 2023-03-13
GH action build image using redhat-actions/buildah-build@v2

eguzki:
### what

* Using `redhat-actions/buildah-build@v2` action to build images. Same as other kuadrant components
* No `unknown on unknown` docker child images created, only `linux on amd64` and `linux on arm64` as expected
* Simplify docker image build. From two step builds workflow (deps + limitador), to one step build workflow.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -24,30 +24,29 @@ jobs:
         id: add-latest-tag
         run: |
           echo "IMG_TAGS=latest" >> $GITHUB_ENV
-
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2        
-      - name: Setup Docker buildx
-        uses: docker/setup-buildx-action@v2
-      - name: Login to Quay.io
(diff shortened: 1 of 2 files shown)


eguzki:
* On amd64, image build times
  * dep build:  `5m 59s`
  * limitador and few deps build: `3m 27s`
* On arm64, image build times :facepalm: 
  * dep build:  `58m 01s`
  * limitador and few deps build: `27m 28s`

eguzki:
After removing the dep build step:
* On amd64, `9m 40s` (same as two step build)
* On arm64, `81m 32s` (~85min took the two step build)

Does the change (two step -> one step) make any sense?

alexsnaps:
Ugh... that's sad. 4 minutes is still a 5% win... wdyt?

eguzki:
Do we need the "feature"? Change the limitador code and "save" docker image build time? 

I do not need that "feature". I leave the final decision to you :)

alexsnaps (approved):
oK then lets merge the two steps and see if/when someone complains

Issue #156, opened 2023-03-13
Get `enable_x_ratelimit_headers: DRAFT_VERSION_03` to work


chirino:
I tried setting the [enable_x_ratelimit_headers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#envoy-v3-api-field-extensions-filters-http-ratelimit-v3-ratelimit-enable-x-ratelimit-headers) option in the envoy rate limiting filter.  But the expected headers were not set in the response.

the filter's configuration looks like: 
```
                  - name: envoy.filters.http.ratelimit
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimit
                      domain: nexodus
                      failure_mode_deny: false
                      timeout: 0.05s
                      enable_x_ratelimit_headers: DRAFT_VERSION_03
                      rate_limit_service:
                        transport_api_version: V3
                        grpc_service:
                          envoy_grpc:
                            cluster_name: ratelimiter
```

The only header I see in the responses is `x-envoy-ratelimited: true`, I was hoping to get headers like `X-RateLimit-Limit: 10, 10;w=1;name="per-ip", 1000;w=3600`

 ## Tasks

 - [x] Add support in Limitador's gRPC
 - [x] Add configuration option to Limitador server
 - [x] Add support in the wasm-shim
 - [x] [Add support to configure in limitador-operator](https://github.com/Kuadrant/limitador-operator/pull/66)

pehala:
Not tested by QE currently.

PR #157, opened 2023-03-19
Adding response headers in the RLS server 

chirino:
so that it implements https://datatracker.ietf.org/doc/id/draft-polli-ratelimit-headers-03.html

See issue #156

### DIFF
--- a/limitador-server/src/envoy_rls/server.rs
+++ b/limitador-server/src/envoy_rls/server.rs
@@ -1,3 +1,4 @@
+use crate::envoy_rls::server::envoy::config::core::v3::HeaderValue;
 use crate::envoy_rls::server::envoy::service::ratelimit::v3::rate_limit_response::Code;
 use crate::envoy_rls::server::envoy::service::ratelimit::v3::rate_limit_service_server::{
     RateLimitService, RateLimitServiceServer,
@@ -6,7 +7,8 @@ use crate::envoy_rls::server::envoy::service::ratelimit::v3::{
     RateLimitRequest, RateLimitResponse,
 };
 use crate::Limiter;
-use std::collections::HashMap;
+use limitador::counter::Counter;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use tonic::{transport, transport::Server, Request, Response, Status};
 
@@ -62,23 +64,29 @@ impl RateLimitService for MyRateLimiter {
             req.hits_addend
         };
 
-        let is_rate_limited_res = match &*self.limiter {
-            Limiter::Blocking(limiter) => {
-                limiter.check_rate_limited_and_update(&namespace, &values, i64::from(hits_addend))
-            }
+        let rate_limited_resp = match &*self.limiter {
+            Limiter::Blocking(limiter) => limiter.check_rate_limited_and_update_getting_counters(
+                &namespace,
+                &values,
+                i64::from(hits_addend),
+            ),
             Limiter::Async(limiter) => {
                 limiter
-                    .check_rate_limited_and_update(&namespace, &values, i64::from(hits_addend))
+                    .check_rate_limited_and_update_getting_counters(
+                        &namespace,
+                        &values,
+                        i64::from(hits_addend),
+                    )
                     .await
             }
         };
 
-        let resp_code = match is_rate_limited_res {
-            Ok(rate_limited) => {
+        let (resp_code, counters) = match rate_limited_resp {
+            Ok((rate_limited, counters)) => {
                 if rate_limited {
-                    Code::OverLimit
+                    (Code::OverLimit, counters)
                 } else {
-                    Code::Ok
+                    (Code::Ok, counters)
                 }
             }
             Err(e) => {
@@ -100,7 +108,7 @@ impl RateLimitService for MyRateLimiter {
             overall_code: resp_code.into(),
             statuses: vec![],
             request_headers_to_add: vec![],
-            response_headers_to_add: vec![],
+            response_headers_to_add: to_response_header(counters),
             raw_body: vec![],
             dynamic_metadata: None,
             quota: None,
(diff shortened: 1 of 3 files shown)


chirino (commented):
this PR still needs work.. but I need a little assistance..

limitador/src/lib.rs
```diff
@@ -375,27 +375,29 @@ impl RateLimiter {
         namespace: &Namespace,
         values: &HashMap<String, String>,
         delta: i64,
-    ) -> Result<bool, LimitadorError> {
+    ) -> Result<(bool, Vec<Counter>, Option<String>), LimitadorError> {
```
alexsnaps, line 378: I think it'd be better to add a new method for that path... At the library level (i.e. what's in `limitador/limitador` being able to chose whether to pay the overhead of being able to get that data should be a "first class citizen"

limitador-server/src/envoy_rls/server.rs
```diff
@@ -110,6 +112,37 @@ impl RateLimitService for MyRateLimiter {
...
+        buf.push_str(format!("{}, {};w={}", limit, limit, window).as_str());
+
+        if let Some(name) = counter.limit().name() {
+            buf.push_str(format!(";name=\"{}\"", name).as_str());
+        }
+        headers.push(HeaderValue { key: "X-RateLimit-Limit".to_string(), value: buf });
+        if let Some(remaining) = counter.remaining() {
+            headers.push(HeaderValue { key: "X-RateLimit-Remaining".to_string(), value: format!("{}", remaining) });
+        }
+        if let Some(duration) = counter.expires_in() {
```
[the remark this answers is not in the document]
alexsnaps: If you look at how the check is performed, i.e. [for redis](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/redis/redis_async.rs#L59), you'll see that the actual value is checked against, then dropped on the floor. 

Probably adding a "seam" that would take a Vec of `&mut Counter` so to update the value in the instance alongside doing the check _and_, again in the case of Redis, query the `TTL` for all these counters and updating the `expires_in` field of each `Counter` would be a first stab at this. This probably explains my other comment about creating an additional method as, depending on the counter storage this will have non neglectable overhead.

We probably could add another script that compounds these operations instead of doing the `MGET` as it is currently done and do the update while doing a `MGET/MTTL` of all the values in one go. I don't know what infinispan provides as options to address this, but a O(2n) would be initially acceptable, as we don't "ship" that anyways. 

To TL;DR I'd go for a naive approach initially and have the support be configurable, so that the overhead isn't imposed on everyone initially.

alexsnaps:
@chirino thanks for looking into this, as a general advise, please run `cargo fmt`, as this will avoid unnecessary formatting changes to your changeset (and avoid the test failure then too). Also running `cargo clippy` will help you avoiding doing common mistakes in Rust and lead you into writing more idiomatic Rust (and again, save from a test failure)... Hope this helps.

limitador/src/lib.rs
chirino: Do you guys have perf tests to measure what the additional overhead would be?

alexsnaps: Yes, have a redis running, and run `cargo bench`

alexsnaps:
> @alexsnaps with these changes the headers added to are showing up in response to the api client.

Yes, but you'd want those on the actual HTTP request, don't you? Or am I misunderstanding this?

chirino:
As a consumer of an API, I want them in the API *responses* (not the requests to the upstream) so that I have an idea of how much API quota I have left and know the limits and time windows that are being applied.   And this is what is occurring now.

alexsnaps:
Ugh, yeah, sorry I meant HTTP response, not request... So it means envoy is actually handling these for us? That's great then. I guess my main concern is that we are still unsure whether we'll keep the protocol as is between envoy (i.e. the wasm-shim) and limitador. I'll look into how this get propagated now then...

alexsnaps:
Right, I trust you... I just don't _know how_ these headers are propagated... And somehow that means that this grpc request from the wasm-shim is treated as a rlp one from envoy's perspective. So if we'd ever change that protocol, that'd mean it'd break this if I understand this right... or say "how I'm guessing this", as again... I know _I don't know_.

chirino:
The ability to add headers is part of the RLP service interface v3.  I've not changed anything there.  See: 
https://github.com/Kuadrant/limitador/blob/main/limitador-server/vendor/protobufs/data-plane-api/envoy/service/ratelimit/v3/rls.proto#L190

alexsnaps:
Could it be you use limitador with Envoy's RL filter? And not our wasm-shim?

chirino:
Sure this kind of response header would be handy in that case too.

limitador/src/lib.rs
```diff
@@ -400,6 +400,39 @@ impl RateLimiter {
...
+            .collect();
+
+        if counters.is_empty() {
+            self.prometheus_metrics.incr_authorized_calls(namespace);
+            return Ok((false, HashSet::new()));
+        }
+
+        let check_result = self.storage.check_and_update(counters.clone(), delta)?;
+
+        let counters = self.storage.load_counters(counters)?;
```
alexsnaps, line 421: This is quite a "big" window for races here. Data will be quite wrong and observable by the user for qualified counters... We should probably add a (hopefully race-free) way to have accurate counters. Also, worth considering the overhead... We currently do have a script [`VALUES_AND_TTLS`](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/redis/scripts.rs#L26-L32) that does get the counter values and associated TTL in a single round trip to Redis. We should consider minimizing the overhead here... From slack:

Alex Snaps
>If that has say over 10% of degradation in latency, then I think we should make this optional in limitador's config...
>(kinda pulled this figure out of thin air tho)

Hiram Chirino
>+1 

So that the overhead of adding the headers is optional and can be (dis|en)abled by the user to...

alexsnaps (commented):
Recorded a few points from the discussions that happened in parallel on Slack, so that if anyone wants to pick this up again, what led to closing this is available for all to see here.

Issue #71, opened 2022-05-26
`InMemoryStore` grows unbounded


alexsnaps:
While we do have a `TtlCache<Counter, i64>`, the `Set` of `Counter` in `limits_for_namespace: RwLock<HashMap<Namespace, HashMap<Limit, HashSet<Counter>>>>` will grow unbounded. The `TtlCache<Counter, i64>`, limited to 1000 entries by default, will evict expired entries (when at capacity I guess?), the `Set` of `Counter`s in our "main" datastructure is unbounded. Not necessarily a big issue, as it "only" will grow for "qualified" (probably not the best name) `Limit`s (i.e. ones that have a counter per some arbitrary variable). 

If we use only a single set of counters (i.e. the one in the main datastructure), we and once we have a non-blocking way of maintaining these, we probably should implement some eviction/expiry mechanism to reclaim memory of these counters. 

<details>

I've debugged this using this test below… while the `rate_limiter.get_counters("foo".to_owned()).unwrap().len()` will return `0`, going deeper in the `InMemoryStore` you'd see that the value of the `<Limit, HashSet<Counter>>` mapping for that `Limit` is a HashSet of size a `1000`. It "merely" gets cleaned up in using the TTL of the cache in: 

```rust
        for counter in self.counters_in_namespace(namespace) { // 1000 counters
            if let Some(counter_val) = self.counters.read().unwrap().get(&counter) { // which are all expired here
                let mut counter_with_val = counter.clone();
                counter_with_val.set_remaining(*counter_val);
                res.insert(counter_with_val);
            }
        }
```

```rust
#[cfg(test)]
mod tests {
    use std::collections::HashMap;
    use std::thread;
    use std::time::Duration;
    use crate::{Limit, RateLimiter};

    #[test]
    fn test_leak() {
        let rate_limiter = RateLimiter::default();
        let nothing: Vec<String> = vec![];
        let limit = Limit::new(
            "foo",
            2,
            1,
            nothing.clone(),
            vec!["id"],
        );
        rate_limiter.add_limit(&limit).unwrap();
        for o in 0..3 {
            for l in 0..3 {
                for i in 0..1000 { // grow the store even bigger, by making this >1000...
                    let mut map = HashMap::new();
                    map.insert("id".to_owned(), i.to_string());
[...]

PR #158, opened 2023-03-21
Adding response headers in the RLS server 

chirino:
so that it implements https://datatracker.ietf.org/doc/id/draft-polli-ratelimit-headers-03.html

See issue #156


chirino:
@alexsnaps could ya review this one?

limitador/src/storage/in_memory.rs
```diff
@@ -31,22 +31,40 @@ impl CounterStorage for InMemoryStorage {
...
-            if !Self::counter_is_within_limits(counter, stored_counters.get(counter), delta) {
-                return Ok(Authorization::Limited(
-                    counter.limit().name().map(|n| n.to_owned()),
-                ));
+        if load_counters {
+            let mut first_limited = None;
+            for counter in counters.iter_mut() {
+                let remaining =
+                    *stored_counters.get(counter).unwrap_or(&counter.max_value()) - delta;
+                counter.set_remaining(remaining);
```
alexsnaps, line 46: We're not updating the `expires_in` duration on this store... Not that I know how we'd do that. `TtlCache` doesn't seem to give us access to `InternalEntry.expiration` 🤔 
But I think we should get rid of it anyways... see #71

limitador/src/storage/mod.rs
```diff
@@ -121,10 +121,12 @@ impl Storage {
 
     pub fn check_and_update(
         &self,
-        counters: HashSet<Counter>,
+        counters: &mut Vec<Counter>,
```
alexsnaps, line 124: Is there a reason you made this a `Vec`? Cause it eventually gets sorted?

alexsnaps (commented):
I had to tweak the bench to actually hit the new code path (loading the counters). I need to rerun it all on my linux machine tho, MacOS is too flaky on my runs, BUT I do seem to barely see any perf degradation... which makes me wonder if we should burn the other path and get rid of that `bool` altogether... wdyt?
Also, do we need to make this configurable? Is there a case where people would NOT want those headers to be present?

limitador/src/storage/in_memory.rs
chirino: what is the 'it' we should get rid of?

limitador/src/storage/mod.rs
chirino: and HashSet does not have a iter_mut() method.

limitador/src/storage/redis/redis_sync.rs
```diff
@@ -157,6 +147,50 @@ impl RedisStorage {
     }
 }
 
+pub fn is_limited(
```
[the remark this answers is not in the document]
chirino: will do.

limitador/src/storage/in_memory.rs
```diff
@@ -31,22 +31,40 @@ impl CounterStorage for InMemoryStorage {
...
+                    return Ok(Authorization::Limited(
+                        counter.limit().name().map(|n| n.to_owned()),
+                    ));
+                }
             }
         }
 
-        for counter in counters {
-            self.insert_or_update_counter(&mut stored_counters, &counter, delta)
+        for counter in counters.iter() {
```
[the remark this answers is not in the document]
chirino: yep.. will fix.

limitador-server/src/envoy_rls/server.rs
```diff
@@ -110,6 +122,54 @@ impl RateLimitService for MyRateLimiter {
...
+            Ordering::Greater
+        }
+    });
+
+    let mut all_limits_text = String::with_capacity(20 * counters.len());
+    counters.iter_mut().for_each(|counter| {
+        all_limits_text
+            .push_str(format!(", {};w={}", counter.max_value(), counter.seconds()).as_str());
+        if let Some(name) = counter.limit().name() {
+            all_limits_text.push_str(format!(";name=\"{}\"", name).as_str());
```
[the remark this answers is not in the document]
chirino: Likely.. the spec BNF does not clarify this very well.  What are the escaping rules?  Maybe I should just translate all `"` chars to something else like `'`

limitador/src/storage/in_memory.rs
alexsnaps: `TtlCache`, but that's not for this PR...

limitador/src/storage/mod.rs
alexsnaps: Ah, yeah of course it doesn't... But given the performance improvement avoiding to dedup as we put them in a Set, I'd say this is fine, users "just" need to make sure they don't provide dups...

chirino: Should we dedup in the counters_that_apply function?

alexsnaps: No, I think the perf gain is worth it, it'd be weird someone feeds in dups anyways...

alexsnaps:
What we now need is a way to configure the server to call into the limiter's `check_rate_limited_and_update_getting_counters` or `check_rate_limited_and_update` and set the headers conditionally as well. Based on my testing, in cases where you have multiple counters per request, this leads to a up to 20% latency degradation.

PR #160, opened 2023-03-25
fix: added integration tests for load_counters=true

chirino:
fixed issue with counters not being properly incremented.

Seems that loading the counters first would change the key names since the remaining and ttl are part of the data used to compute the key name.  We now use the key names computed before loading the counters to increment the counters.  Better fix might be to have a better key_for_counter func.


alexsnaps:
Fixed with #161

PR #163, opened 2023-03-30
Add option to manually trigger job

alexsnaps:


.github/workflows/build-image.yaml
```diff
@@ -2,6 +2,7 @@
 name: Build Image
 
 on:
+  workflow_dispatch:
```
eguzki, line 5: empty? `inputs` is not required?

alexsnaps: I don't think so, and I really don't need one... Just have no idea why the latest merge didn't trigger this workflow?!

alexsnaps:
image
👌

PR #168, opened 2023-04-27
Move to UBI8 parent image

gsaslis:
This will help with producing a product build for limitador, and makes many things easier such as rebuilds on Common Vulnerability and Exposure (CVE) patches, (e.g. Container Health Index (CHI) - the ‘A’ / ‘B’ / ‘C’ grades shown on Red Hat Container Catalog (RHCC) ), cross-product dependencies, etc.

Notes:
* slightly different package names in ubi8, but this shouldn't be introducing anything new.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,24 @@
 # Build Stage
 # ------------------------------------------------------------------------------
 
-FROM alpine:3.16 as limitador-build
+FROM registry.access.redhat.com/ubi8/ubi:8.7 as limitador-build
 ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
 
 ARG RUSTC_VERSION=1.67.1
-RUN apk update \
-    && apk upgrade \
-    && apk add build-base binutils-gold openssl3-dev protoc protobuf-dev curl git \
-    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path --profile minimal --default-toolchain ${RUSTC_VERSION} -c rustfmt -y
+
+# the powertools repo is required for protobuf-c and protobuf-devel
+RUN dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install \
+      http://mirror.centos.org/centos/8-stream/BaseOS/`arch`/os/Packages/centos-gpg-keys-8-6.el8.noarch.rpm \
+      http://mirror.centos.org/centos/8-stream/BaseOS/`arch`/os/Packages/centos-stream-repos-8-6.el8.noarch.rpm \
+ && dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install epel-release \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,14 +2,24 @@
...
 
 ARG RUSTC_VERSION=1.67.1
-RUN apk update \
-    && apk upgrade \
-    && apk add build-base binutils-gold openssl3-dev protoc protobuf-dev curl git \
-    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path --profile minimal --default-toolchain ${RUSTC_VERSION} -c rustfmt -y
+
+# the powertools repo is required for protobuf-c and protobuf-devel
+RUN dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install \
+      http://mirror.centos.org/centos/8-stream/BaseOS/`arch`/os/Packages/centos-gpg-keys-8-6.el8.noarch.rpm \
```
eguzki, line 12: centos repos?  I thought it reached EOL

Would it work compiling in alpine and run in ubi8?

Dockerfile
```diff
@@ -26,12 +36,16 @@ RUN source $HOME/.cargo/env \
 # Run Stage
 # ------------------------------------------------------------------------------
 
-FROM alpine:3.16
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7
```
eguzki, line 39: ubi9?

gsaslis: I'm not familiar with all the ubi9 changes related to product builds, so I'd rather not.

Dockerfile
gsaslis: centos stream is fully supported until at least next year: https://wiki.centos.org/About/Product

alexsnaps: >Would it work compiling in alpine and run in ubi8?

Why would you prefer that? Just to avoid the CentOS repo?

eguzki: Well, the build stage seems now harder to maintain. With alpine, it was easier to maintain (not simpler though).

Not a big deal, though. LGTM

eguzki: And yes, CENTOS in my mind is an echo of the PAST

gsaslis: > Well, the build stage seems now harder to maintain. With alpine, it was easier to maintain (not simpler though).

definitely harder / more obscure to maintain. 👍🏼 

the alternative would be to build protobuf from source though, which I guess is worse... With that assumption, I'll go ahead and resolve - but please reopen if you prefer us to move in that direction.

gsaslis:
guess this is ready to merge ?

eguzki (approved):
As @alexsnaps pointed out offline quite rightly, the runtime base image should be aligned with the building base image to avoid issues with dynamic linking. Most concerning one: openssl

alexsnaps:
Thanks for the PR @gsaslis

PR #167, opened 2023-04-19
Sled serde

alexsnaps:
Still needs: 

 - [ ] Fix infinispan module…

PR #170, opened 2023-05-02
🔥 Sled and use RocksDB instead

alexsnaps:
This builds on and supersedes both #146 & #167 … It replaces [sled](https://crates.io/crates/sled) as the underlying storage for local persistence with [RocksDB](https://rocksdb.org/) using the [Rust bindings](https://crates.io/crates/rocksdb)… which both were extensively battle tested. 

If we think this is indeed a better choice, it'd probably be worthwhile `git rebase -i` from HEAD of main, squashing all the work from the other PRs… I kept them as is for now, hoping this makes reviewing somewhat easier.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
 ARG RUSTC_VERSION=1.67.1
 RUN apk update \
     && apk upgrade \
-    && apk add build-base binutils-gold openssl3-dev protoc protobuf-dev curl git \
+    && apk add build-base binutils-gold openssl3-dev protoc protobuf-dev curl git linux-headers clang \
     && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path --profile minimal --default-toolchain ${RUSTC_VERSION} -c rustfmt -y
 
 WORKDIR /usr/src/limitador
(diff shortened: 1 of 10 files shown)


eguzki:
> hoping this makes reviewing somewhat easier.

It does, thanks for that

eguzki:
Looking good. I would go ahead and try to add the `disk persistence` feature, at least in `Tech Preview`

Dockerfile
```diff
@@ -8,7 +8,7 @@ ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
 ARG RUSTC_VERSION=1.67.1
 RUN apk update \
     && apk upgrade \
-    && apk add build-base binutils-gold openssl3-dev protoc protobuf-dev curl git \
+    && apk add build-base binutils-gold openssl3-dev protoc protobuf-dev curl git linux-headers clang \
```
eguzki, line 11: there may be issues with the new ubi8 based images https://github.com/Kuadrant/limitador/pull/168

alexsnaps: Totally will be...

alexsnaps:
> Have you considered using the [Compaction Filter](https://github.com/facebook/rocksdb/wiki/RocksDB-Overview#compaction-filter) to implement the TTL??

I did, the issue is that 

> Have you considered using the [Compaction Filter](https://github.com/facebook/rocksdb/wiki/RocksDB-Overview#compaction-filter) to implement the TTL??

I do when disk optimized: https://github.com/Kuadrant/limitador/pull/170/files#diff-b41633d05540bf2734d97b29f9dab57f411c275100114fa933f6b0582cd04abaR124-R135

alexsnaps:
Closed in favor of #178

PR #173, opened 2023-05-08
build image workflow: smoke test

eguzki:


eguzki:
I deliberately changed the Dockerfile to see if the smoke test rings the bell

```diff
diff --git a/Dockerfile b/Dockerfile
index 4160fb1..b7195ec 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,7 +48,7 @@ RUN groupadd -g 1000 limitador \
     && useradd -u 1000 -g limitador -s /bin/sh -m -d /home/limitador limitador
 
 WORKDIR /home/limitador/bin/
-ENV PATH="/home/limitador/bin:${PATH}"
+ENV PATH="/home/limitador/bin3:${PATH}"
 
 COPY --from=limitador-build /usr/src/limitador/limitador-server/examples/limits.yaml ../
 COPY --from=limitador-build /usr/src/limitador/target/release/limitador-server ./limitador-server
``` 

And, indeed, the smoke tests failed https://github.com/Kuadrant/limitador/actions/runs/4925559127/jobs/8799930487

```
Run podman run --rm -t limitador:bf52d1e1684bf9a7676f12c5c066bc794a0e5eb0 limitador-server --help
Error: executable file `limitador-server` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
Error: Process completed with exit code 127.
```

eguzki:
@Kuadrant/engineering ready for review!

alexsnaps (approved):
I guess this is different from [e2e](https://github.com/Kuadrant/limitador/blob/main/.github/workflows/e2e.yaml) in that it would fail pushing a broken image, i.e. one where something went wrong probably wrt dynamic linking, to quay.io, is that right?

eguzki:
> I guess this is different from [e2e](https://github.com/Kuadrant/limitador/blob/main/.github/workflows/e2e.yaml) in that it would fail pushing a broken image, i.e. one where something went wrong probably wrt dynamic linking, to quay.io, is that right?

Yes. Here "broken image" means the simple test `limitador-server --help` returns not `0`. e2e tests would also fail, but this smoke test tells precisely that the limitador image does not start, i.e. that the binary `limitador-server` cannot be executed.

PR #180, opened 2023-05-19
Bench: Actually create different limits for each scenario

alexsnaps:
So this is going to hurt:

We always created the bench "wrong", or rather we created dups of one identical `Limit` in each namespace, when it was more than `1` (i.e. `10` & `50`). Sadly this meant that Limitador would only ever do _one_ limit  (with more variables and conditions tho)... So this fixes that, and as such exposes the `O(n)` problem I mentioned in our refinement call...

#### In memory:

image

#### On disk:

image

#### With Redis:

image

As expected, Memory is the worst: `O(n)`, while Redis, which does an `MGET` is the least impacted. While the operation is still `O(n)` within Redis, the network overhead ans (de)serialization dominates the latency here…

### DIFF
--- a/limitador/benches/bench.rs
+++ b/limitador/benches/bench.rs
@@ -280,11 +280,11 @@ fn generate_test_data(
     for idx_namespace in 0..scenario.n_namespaces {
         let namespace = idx_namespace.to_string();
 
-        for _ in 0..scenario.n_limits_per_ns {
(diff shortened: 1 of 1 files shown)


eguzki:
I am still surprised with the disk performance. I was expecting much worse performance. I guess actual numbers depend on the disk properties.

Issue #183, opened 2023-06-01
Avoid doing too many hash lookups with for a single counter


alexsnaps:
Currently, we store Counters twice when using the `InMemoryStorage`, once in the `limits_for_namespace` within a `HashSet` and in the counters field, the `TtlCache`. 
For that to happen, we need to calculate the hash of the `Counter`, which needs the hash to the `Limit`, which itself is also needed to find the right bucket in our `HashMap<Limit, HashSet<Counter>>`. All in all most of the time is effectively spent calculating hash values to store our counters. 
When [checking for a counter](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/in_memory.rs#LL32C16-L32C16), rather than storing [them twice](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/in_memory.rs#L177-L184), we want to store them only once. We need the entry in the `TtlCache`, because counters expire. The proposal is to use an approach comparable to what the `RocksDbStorage` uses and have the expiry being handled explicitly by our own stored counter values. Currently, because the `TtlCache` also doesn’t allow querying the TTL of an entry, the Rate Limit headers won’t expose the time window when counters are stored in-memory. Changing this would allow us to fix that as well.
Finally, the looking at introducing proper key types used to lookup actual `Counter` and/or `Limit` instances in our data structures would be a way of avoiding doing double calculations of hash values, e.g. when querying for a `Counter` for `Limit` (as the `Counter` hash function hashes the `Limit`).

alexsnaps:
@didierofrivia I think you stopped working on this, or didn't you?

PR #186, opened 2023-06-12
 Improving perf when using InMemory storage

didierofrivia:
This is also part of https://github.com/Kuadrant/limitador/issues/183

### Criterion performance results for In Memory storage, check_rate_limited_and_update function
Screenshot 2023-07-03 at 16 59 08

### 10 namespaces with 10 limits each with 10 conditions and 10 variables
Screenshot 2023-07-03 at 16 58 54

### 10 namespaces with 50 limits each with 10 conditions and 10 variables

Screenshot 2023-07-03 at 16 58 40

Screenshot 2023-07-03 at 18 05 27


limitador/src/storage/in_memory.rs
```diff
@@ -1,31 +1,98 @@
...
+    }
+}
 
 pub struct InMemoryStorage {
-    limits_for_namespace: RwLock<HashMap<Namespace, HashMap<Limit, HashSet<Counter>>>>,
-    counters: RwLock<TtlCache<Counter, i64>>,
+    //TODO: This is a bit ugly, to address in the future
+    #[allow(clippy::type_complexity)]
+    limits_for_namespace:
+        RwLock<HashMap<Namespace, HashMap<Limit, HashMap<CounterKey, ExpiringValue>>>>,
```
alexsnaps, line 57: Do you think you'd eventually make this simpler? Otherwise a type alias might be a good idea already, wdyt?

limitador/src/storage/in_memory.rs
```diff
@@ -35,36 +102,88 @@ impl CounterStorage for InMemoryStorage {
...
+                        }
+                    }
+                } else if let Some(limited) = process_counter(counter, 0, delta) {
+                    if !load_counters {
+                        return Ok(limited);
+                    }
+                }
+            } else if let Some(limited) = process_counter(counter, 0, delta) {
+                if !load_counters {
+                    return Ok(limited);
```
alexsnaps, line 159: I think we could easily avoid this duplication here, and only do this once, maybe after this entire logic…

alexsnaps (changes requested):
This looks real good to me, if we could at least clean up the repetition there…

PR #191, opened 2023-07-18
AtomicExpiringValue

didierofrivia:


limitador/src/storage/atomic_expiring_value.rs
```diff
@@ -0,0 +1,135 @@
...
+            return 0;
+        }
+        self.value.load(Ordering::SeqCst)
+    }
+
+    pub fn value(&self) -> i64 {
+        self.value_at(SystemTime::now())
+    }
+
+    pub fn update(&self, delta: i64, ttl: u64, when: SystemTime) {
```
alexsnaps, line 39: With these changes, when we get rid of the RwLock guarding concurrent updates, we will want to return the counter's `value` here and probably recheck it against the Limit's ... We probably might want to discuss this at some point.

limitador/src/storage/atomic_expiring_value.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+impl Default for AtomicExpiringValue {
+    fn default() -> Self {
+        AtomicExpiringValue {
+            value: AtomicI64::new(0),
+            expiry: AtomicU64::new(
+                UNIX_EPOCH
+                    .duration_since(UNIX_EPOCH)
```
alexsnaps, line 76: Isn't this always `0`?

limitador/src/storage/in_memory.rs
```diff
@@ -123,56 +125,44 @@ impl CounterStorage for InMemoryStorage {
...
+            limits_by_namespace
+                .entry(counter.limit().namespace().clone())
+                .or_insert_with(HashMap::new)
+                .entry(counter.limit().clone())
+                .or_insert_with(HashMap::new)
+                .entry(counter.into())
+                .or_insert_with(|| {
+                    AtomicExpiringValue::new(0, now + Duration::from_secs(counter.seconds()))
+                });
+        }
```
alexsnaps, line 139: It'd be great to see if we can trigger this on loading a (newer) Limit file. So that this structure is always in sync with the Limits held in memory. This _would need to be atomic_ with the loading of the newer Limits

limitador/src/storage/in_memory.rs
```diff
@@ -123,56 +125,44 @@ impl CounterStorage for InMemoryStorage {
...
-                }
-                Entry::Occupied(mut o) => {
-                    o.get_mut().update_mut(delta, counter.seconds(), now);
-                }
-            }
-        }
+        // Update counters
+        counter_values_to_update
+            .iter()
+            .for_each(|(v, ttl)| v.update(delta, *ttl, now));
```
alexsnaps, line 165: This is only not racy right now because of the WriteLock acquired above.

limitador/src/storage/in_memory.rs
alexsnaps: And the value should be `AtomicExpiringValue::default()`... so that the value present _is expired_ already. Otherwise the clock towards expiry would start ticking when the limits are loaded, while the first request might come 30 seconds later... Not a big deal, but it'd change the current behavior.

alexsnaps (approved):

Issue #169, opened 2023-04-28
Setup CodeCov integration


eguzki:
https://app.codecov.io/gh/Kuadrant

PR #192, opened 2023-07-24
workflow: add codecov

KevFan:
* Add coverage
  * Test and generate code coverage using https://github.com/taiki-e/cargo-llvm-cov 

Closes: https://github.com/Kuadrant/limitador/issues/169


alexsnaps (approved):
Let's keep this as is for now, but if we get some issues with the tests with nightly (we shouldn't but... famous last words) - we may want to run tests on stable (without coverage) too. Also, leaving [this](https://doc.rust-lang.org/rustc/instrument-coverage.html) for reference...

PR #193, opened 2023-07-26
[fix] Adding "/" and "-" chars for the new condition format

didierofrivia:
This PR addresses the need to change the new condition format introduced [RLP V2 in Kuadrant](https://github.com/Kuadrant/kuadrant-operator/pull/218)

### DIFF
--- a/limitador/src/limit.rs
+++ b/limitador/src/limit.rs
@@ -613,7 +613,7 @@ mod conditions {
 
         fn valid_id_char(&mut self) -> bool {
             let char = self.input[self.pos];
-            char.is_alphanumeric() || char == '.' || char == '_'
+            char.is_alphanumeric() || char == '.' || char == '_' || char == '-' || char == '/'
         }
 
         fn scan_string(&mut self, until: char) -> Result<Token, SyntaxError> {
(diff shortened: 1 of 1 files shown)


guicassolato (approved):

limitador/src/limit.rs
```diff
@@ -613,7 +613,7 @@ mod conditions {
 
         fn valid_id_char(&mut self) -> bool {
             let char = self.input[self.pos];
-            char.is_alphanumeric() || char == '.' || char == '_'
+            char.is_alphanumeric() || char == '.' || char == '_' || char == '-' || char == '/'
```
alexsnaps, line 616: So the trouble I have with this change is that it muddies the waters... What if it's to be an expression now? Like `http.response_code - 12` ? The white spaces matter now? 
That being said I don't think we'd ever support that (very example if only 🤣 ). That being said I already hated having the `.` in there... tho I thought you could have argued it remained an identifier for "something" to resolve... 

So anyways, I won't necessarily block this from happening, but I wonder if sanitizing the variable names/identifiers wouldn't be a slightly more desirable.

PR #194, opened 2023-07-27
fix: codecov badge using wrong branch

KevFan:
* Codecov badge was using `master` instead of `main` so the code coverage percentage was not shown from inital work in  https://github.com/Kuadrant/limitador/pull/192


adam-cattermole (approved):
Looks good

PR #195, opened 2023-07-28
Store qualified counters in dedicated capacity limited cache

alexsnaps:
@didierofrivia here is what I am thinking...

### DIFF
--- a/limitador-server/src/envoy_rls/server.rs
+++ b/limitador-server/src/envoy_rls/server.rs
@@ -246,7 +246,7 @@ mod tests {
             vec!["app_id"],
         );
 
-        let limiter = RateLimiter::default();
+        let limiter = RateLimiter::new(10_000);
         limiter.add_limit(limit);
 
         let rate_limiter = MyRateLimiter::new(
@@ -366,7 +366,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_takes_into_account_all_the_descriptors() {
-        let limiter = RateLimiter::default();
+        let limiter = RateLimiter::new(10_000);
 
         let namespace = "test_namespace";
 
@@ -434,7 +434,7 @@ mod tests {
         let namespace = "test_namespace";
         let limit = Limit::new(namespace, 10, 60, vec!["x == '1'"], vec!["y"]);
 
-        let limiter = RateLimiter::default();
+        let limiter = RateLimiter::new(10_000);
         limiter.add_limit(limit);
 
         let rate_limiter = MyRateLimiter::new(
@@ -499,7 +499,7 @@ mod tests {
         let namespace = "test_namespace";
         let limit = Limit::new(namespace, 1, 60, vec!["x == '1'"], vec!["y"]);
 
-        let limiter = RateLimiter::default();
+        let limiter = RateLimiter::new(10_000);
         limiter.add_limit(limit);
 
         let rate_limiter = MyRateLimiter::new(
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -7,14 +7,15 @@ extern crate clap;
 #[cfg(feature = "infinispan")]
 use crate::config::InfinispanStorageConfiguration;
 use crate::config::{
-    Configuration, DiskStorageConfiguration, RedisStorageCacheConfiguration,
-    RedisStorageConfiguration, StorageConfiguration,
+    Configuration, DiskStorageConfiguration, InMemoryStorageConfiguration,
+    RedisStorageCacheConfiguration, RedisStorageConfiguration, StorageConfiguration,
 };
 use crate::envoy_rls::server::{run_envoy_rls_server, RateLimitHeaders};
 use crate::http_api::server::run_http_server;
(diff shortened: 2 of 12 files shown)


alexsnaps:
These are all the changes... By default, this now _tries_ to use 70% of the available memory at server startup to store "qualifies counters", it'll emit a `WARN`ing level log if you don't explicitly set the cache's size using `-c`, like so:
```
[2023-08-04T14:51:22Z WARN  limitador_server] No cache size provided, aiming at 70% of 6019MB, i.e. 17259155 entries
```

limitador-server/src/envoy_rls/server.rs
```diff
@@ -246,7 +246,7 @@ mod tests {
             vec!["app_id"],
         );
 
-        let limiter = RateLimiter::default();
+        let limiter = RateLimiter::new(10_000);
```
didierofrivia, line 249: I'm missing how this work, but I don't understand why is this change needed? Shouldn't invoke the `default` set the expected cache_size?

limitador-server/src/main.rs
```diff
@@ -785,10 +801,26 @@ fn storage_config_from_env() -> Result<StorageConfiguration, ()> {
...
+        _ => Ok(StorageConfiguration::InMemory(
+            InMemoryStorageConfiguration { cache_size: None },
+        )),
     }
 }
 
+fn guess_cache_size() -> Option<u64> {
+    let sys = System::new_with_specifics(RefreshKind::new().with_memory());
+    let free_mem = sys.available_memory();
+    let memory = free_mem as f64 * 0.7;
```
didierofrivia, line 813: isn't the 70% too much (or too little xD)

limitador/src/lib.rs
```diff
@@ -306,9 +309,9 @@ impl AsyncRateLimiterBuilder {
 }
 
 impl RateLimiter {
-    pub fn new() -> Self {
+    pub fn new(cache_size: u64) -> Self {
```
didierofrivia, line 312: Ok, I kind of understand now why we need the cache size and can't use the default I asked before..

alexsnaps: Yes... there needs to be a value provided (and see the other comment on why I did it this way)

limitador/src/storage/in_memory.rs
```diff
@@ -2,95 +2,78 @@ use crate::counter::Counter;
...
-                if let Some(expiring_value) = counters.get(&counter.into()) {
-                    value = expiring_value.value();
-                }
+
+        if counter.is_qualified() {
+            if let Some(counter) = self.qualified_counters.get(counter) {
+                value = counter.value();
+            }
+        } else if let Some(limits) = limits_by_namespace.get(counter.limit().namespace()) {
+            if let Some(counter) = limits.get(counter.limit()) {
```
didierofrivia, line 30: what we are finding here is the AtomicExpiringValue, do you think it makes sense to be explicit and rename `counter` with something like `expiring_value` ? I also understand it's a _counter_ in the end

limitador-server/src/main.rs
alexsnaps: Well... this math is "wrong", the size of the entries don't account for the overhead of the memory required by the cache itself... but most importantly don't account for the size of the `variables` data that's on the heap... i.e. all the entries of `name/value` pairs... we need some overhead for that.
Finally the "rest" also needs memory, i.e. the actix & tonic & all other plumbing (if only storing the `Limit`s themselves... So for tiny amounts of `sys.available_memory()` 70% might actually be way too eager, while for large ones, probably on the overly cautious end... Which is why this _will trigger a `WARN` level log_.

alexsnaps:
> Left some comments, nothing blocking TBH, feel free to address them if it makes sense or just comment :)

I don't think anything _should_ be addressed, but feel free to disagree... One thing which _might_ be useful is setting a hard limit on what we _know_ would be limitador's memory requirement. But we don't as of now... I could open an issue for that so that we at some point measure and address that. But right now, 70% isn't any better or worse than the previously hardcoded value... it's as bad, just differently. But at least, now, someone can set it to something that makes sense to their use-case.

limitador/src/storage/atomic_expiring_value.rs
```diff
@@ -141,4 +141,11 @@ mod tests {
         });
         assert!([2i64, 3i64].contains(&atomic_expiring_value.value.load(Ordering::SeqCst)));
     }
+
+    #[test]
+    fn size_of_struct() {
+        // This is ugly, but we don't have access to `AtomicExpiringValue` in the server,
```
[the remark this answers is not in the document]
alexsnaps: No. That's the actual size, i.e. `2 * size_of::<u64>() == 2 * 8 == 16`
This test is just here because the value is hardcoded in the server's `main.rs`, as I can't reference `AtomicExpiringValue` from there… So, if we ever happen to make this bigger (or smaller), this test will fail as a reminded to update the value in [here](https://github.com/Kuadrant/limitador/pull/195/files#diff-77c4c3afe1e75f3f1602389f5895694a70395b1a7450ad7776945631b4bd28e9R815)

Issue #182, opened 2023-06-01
Data Model 


alexsnaps:
# The current data model

The data model resides in the limitador crate source and is broken down into these few structs: [`Namespace`](https://github.com/Kuadrant/limitador/blob/main/limitador/src/limit.rs#L32), [`Limit`](https://github.com/Kuadrant/limitador/blob/main/limitador/src/limit.rs#L53-L67) and finally [`Counter`](https://github.com/Kuadrant/limitador/blob/main/limitador/src/counter.rs#L8-L18). 
 - `Namespace` is just a type alias for `String `and mostly represents a hostname for which limits will be applied
 - `Limit` is the representation of the definition of a limit: its targeted Namespace, the maximum value for a given time window and a set of conditions. Additionally counters can be qualified using “variables”, e.g. a counter per user identifier. A Limit can be “named” mostly for tracing/debugging purposes.
 - `Counter`s are the actual counters for a limit, e.g. for a given user identifier. It also holds the time-to-live (TTL) for the counter, e.g. the counter should be reset in 32 seconds as the time window would have expired by then.

Interestingly these types fulfill different roles and are closer to [Data Transfer Objects](https://martinfowler.com/eaaCatalog/dataTransferObject.html) (DTOs) than an actual data model. Counters are mostly used as a lookup mechanism to their actual values. In the memory store, the actual values are held in the [InMemoryStorage.counters’ TtlCache](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/in_memory.rs#L12), that cache is also responsible for applying the correct TTL. Redis is responsible for the same TTL enforcement and stores the values as plain integers in a Hash, keyed by Counter. Limits are too mostly used to look Counters up. They are also used as DTOs for the REST API exposed by the server, e.g. when [querying for known counters](https://github.com/Kuadrant/limitador/blob/main/limitador-server/src/http_api/server.rs#L70-L90).

*Note:* The [RocksDbStorage](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/disk/rocksdb_storage.rs) is slightly different as the underlying RocksDB doesn’t have built-in TTL support. That feature was built in an [ExpiringValue](https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/disk/expiring_value.rs) type that handles the expiry explicitly, while holding the current counter’s value.

eguzki:
@alexsnaps is this still in the scope of v1?

PR #199, opened 2023-08-09
Normalize counter values

didierofrivia:
This PR is part of the many changes in the data model, tracked by https://github.com/Kuadrant/limitador/issues/182

There's no relevant change in performance, I've included a snapshot for reference:
Screenshot 2023-08-09 at 18 20 06

### DIFF
--- a/limitador/src/storage/in_memory.rs
+++ b/limitador/src/storage/in_memory.rs
@@ -96,7 +96,7 @@ impl CounterStorage for InMemoryStorage {
         delta: i64,
         load_counters: bool,
     ) -> Result<Authorization, StorageErr> {
-        let mut limits_by_namespace = self.limits_for_namespace.write().unwrap();
+        let limits_by_namespace = self.limits_for_namespace.write().unwrap();
         let mut first_limited = None;
         let mut counter_values_to_update: Vec<(&AtomicExpiringValue, u64)> = Vec::new();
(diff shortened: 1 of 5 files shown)


limitador/src/storage/in_memory.rs
```diff
@@ -96,7 +96,7 @@ impl CounterStorage for InMemoryStorage {
         delta: i64,
         load_counters: bool,
     ) -> Result<Authorization, StorageErr> {
-        let mut limits_by_namespace = self.limits_for_namespace.write().unwrap();
+        let limits_by_namespace = self.limits_for_namespace.read().unwrap();
```
alexsnaps, line 99: This change introduces a race condition when we eventually update the counters [here](https://github.com/Kuadrant/limitador/pull/199/commits/9711f6338273ad4d44b47b6b8ee29e444b81d49f#diff-ebd607498f8416f1e94e7592efd8580d8dd6798df308dee8f7f7f7f78fe73c18R165-R173). By the time we get to our call `.update` on our `AtomicExpiringValue` references, their value could have already been updated by another thread. This is why we made `.update(...) -> u64`, i.e. return the value _at that updating moment_. But we now need to recheck we didn't pass the threshold... worst, if we did on say the second counter in our vec, what do we do with the increment to the first one?

So, tl;dr, I'm thinking lets leave grabbing the write lock for now and address the concurrency problem in a subsequent PR?

alexsnaps (approved):

PR #213, opened 2023-09-20
coverage test only on main branch

eguzki:
There is an incompatibility between nightly toolchain and [time](https://crates.io/crates/time) v0.3.28. This makes coverage measurement a broken task until new `time` crate is released. The [fix](https://github.com/time-rs/time/pull/622) has already been merged but a new crate has not been release as of today.

This PR runs coverage measurement only on `main` branch, which will keep failing unfortunately. But at least it will not block all the PR's. 

When the new `time` crate is released, the coverage measurement can be brought back to PR's if we want to check coverage increase/decrease due to the PR changes.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,64 @@
+---
+name: Coverage
+
+on:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  check:
+    name: Check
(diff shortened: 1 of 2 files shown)


adam-cattermole (approved):
Nice changes, they look very reasonable from my point of view

alexsnaps (approved):
🙏 Thank you!!!

PR #221, opened 2023-10-19
refactor: use actions-rust-lang/setup-rust-toolchain@v1 & docker/setup-buildx-action@v3 for e2e

KevFan:
* actions-rust-lang/setup-rust-toolchain@v1 incorporates using github action cache. Subsequent builds will use cache if there has been no dependency changes 
  * https://github.com/actions-rust-lang/setup-rust-toolchain
  * https://github.com/Swatinem/rust-cache
* Use docker/setup-buildx-action@v3 to cache docker layers to github action cache. Subsequent builds will use cache if layers have not been invalidated by changes
  * https://github.com/docker/setup-buildx-action
  * https://github.com/docker/build-push-action
* update .dockerignore to include doc related files so that layer cache is not invalidated by changes to these files

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -13,19 +13,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            limitador-testing
+          tags: |
+            type=raw,value=latest
       - name: Build Limitador Image
         id: build-image
-        uses: redhat-actions/buildah-build@v2
+        uses: docker/build-push-action@v5
         with:
-          image: limitador-testing
-          tags: latest
-          dockerfiles: |
-            ./Dockerfile
+          context: .
+          tags: ${{ steps.meta.outputs.tags }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          load: true
       - name: Create k8s Kind Cluster
-        uses: helm/kind-action@v1.1.0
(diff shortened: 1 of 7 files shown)


.github/workflows/e2e.yaml
```diff
@@ -13,19 +13,30 @@ jobs:
...
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            limitador-testing
+          tags: |
+            type=raw,value=latest
       - name: Build Limitador Image
         id: build-image
-        uses: redhat-actions/buildah-build@v2
+        uses: docker/build-push-action@v5
```
KevFan, line 29: Similar changes were **not** done to https://github.com/Kuadrant/limitador/blob/38f55ca1c96246b6532351207453efbf63438c33/.github/workflows/build-image.yaml#L2 due to issues mentioned by https://github.com/Kuadrant/limitador/pull/215.

Additionally as we pass the `GITHUB_SHA` build arg to the dockerfile, we always compile and build on each push as the SHA is always different, so we won't pull from cache which is the most time intensive layer
https://github.com/Kuadrant/limitador/blob/38f55ca1c96246b6532351207453efbf63438c33/.github/workflows/build-image.yaml#L42-L43
https://github.com/Kuadrant/limitador/blob/38f55ca1c96246b6532351207453efbf63438c33/Dockerfile#L26

.github/workflows/e2e.yaml
```diff
@@ -34,8 +45,7 @@ jobs:
           kubectl cluster-info dump
       - name: Load limitador docker image
         run: |
-          podman save -o image.tar ${{ steps.build-image.outputs.image }}:${{ steps.build-image.outputs.tags }}
-          kind load image-archive --name limitador-local image.tar
+          kind load docker-image ${{ steps.meta.outputs.tags }} -n limitador-local
```
eguzki, line 48: kind loads from docker image? Great!

eguzki (approved):

PR #222, opened 2023-10-24
refactor: cross compile for arm64 image build

KevFan:
Building arm64 image via QEMU takes over 4 hours on github actions which makes the development / release process less than ideal.

Building this image via cross compile decreases this to ~20mins
* Changes are based on:
  * https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners
  * https://docs.docker.com/build/building/multi-platform/#cross-compilation
  * https://github.com/Kuadrant/limitador/pull/214
  * https://github.com/Kuadrant/limitador/pull/215

On Mac M1, tested this with:
- amd64 - `docker pull quay.io/kuadrant/limitador:quicker-workflow --platform=linux/amd64`

image

- arm64 - `docker pull quay.io/kuadrant/limitador:quicker-workflow`

image

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -8,7 +8,6 @@ on:
     tags: ['*']
 
 env:
-  IMG_TAGS: ${{ github.sha }}
   IMG_REGISTRY_HOST: quay.io
   IMG_REGISTRY_ORG: kuadrant
   MAIN_BRANCH_NAME: main
@@ -17,43 +16,96 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - dockerfile: Dockerfile
+            platform: linux/amd64
+          - dockerfile: Dockerfile.aarch64
+            platform: linux/arm64
     steps:
       - name: Check out code
         uses: actions/checkout@v4
-      - name: Add latest tag
-        if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
-        run: |
-          echo "IMG_TAGS=${{ env.IMG_TAGS }} latest" >> $GITHUB_ENV
-      - name: Add branch name tag
-        if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
-        run: |
-          echo "IMG_TAGS=${{ env.IMG_TAGS }} ${{ github.ref_name }}" >> $GITHUB_ENV
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
-      - name: Build Image
-        id: build-image
-        uses: redhat-actions/buildah-build@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
         with:
-          image: limitador
-          tags: ${{ env.IMG_TAGS }}
-          platforms: linux/amd64,linux/arm64
-          dockerfiles: |
-            ./Dockerfile
-          build-args: |
-            GITHUB_SHA=${{ github.sha }}
-      - name: Smoke Test
-        run: |
-          podman run --rm -t ${{ steps.build-image.outputs.image }}:${{ github.sha }} limitador-server --help
-      - name: Push Image
-        if: ${{ !env.ACT }}
-        id: push-to-quay
-        uses: redhat-actions/push-to-registry@v2
+          images: |
+            ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador
+      - name: Login to container registry
+        uses: docker/login-action@v2
         with:
-          image: ${{ steps.build-image.outputs.image }}
-          tags: ${{ steps.build-image.outputs.tags }}
-          registry: ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}
           username: ${{ secrets.IMG_REGISTRY_USERNAME }}
           password: ${{ secrets.IMG_REGISTRY_TOKEN }}
-      - name: Print Image URL
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Build and push by digest
+        id: build
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            GITHUB_SHA=${{ github.sha }}
+          cache-from: type=gha,scope=${{ matrix.platform == 'linux/amd64' && 'build-amd' || 'build-arm'}}
+          cache-to: type=gha,mode=max,scope=${{ matrix.platform == 'linux/amd64' && 'build-amd' || 'build-arm'}}
+          outputs: type=image,name=${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador,push-by-digest=true,name-canonical=true,push=true
+          file: ${{ matrix.dockerfile }}
+          platforms: |
+            ${{ matrix.platform }}
+          provenance: false
+      - name: Export digest
         run: |
-          echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
+          mkdir -p /tmp/digests
+          digest="${{ steps.build.outputs.digest }}"
+          touch "/tmp/digests/${digest#sha256:}"
+      - name: Upload digest
+        uses: actions/upload-artifact@v3
+        with:
+          name: digests
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+  merge:
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    steps:
+      - name: Download digests
+        uses: actions/download-artifact@v3
+        with:
+          name: digests
+          path: /tmp/digests
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador
+          tags: |
+            # type=raw,value=${{ github.sha }}
+            # set latest tag for main branch
+            type=raw,value=latest,enable=${{ github.ref_name == env.MAIN_BRANCH_NAME }}
+            # set ref name tag for non-main branches
+            type=raw,value=${{ github.ref_name }},enable=${{ github.ref_name != env.MAIN_BRANCH_NAME }}
+      - name: Login to container registry
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.IMG_REGISTRY_USERNAME }}
+          password: ${{ secrets.IMG_REGISTRY_TOKEN }}
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Create manifest list and push
+        working-directory: /tmp/digests
+        run: |
+          docker buildx imagetools create $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
+            $(printf '${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador@sha256:%s ' *)
+      - name: Inspect image
+        run: |
+          docker buildx imagetools inspect ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador:${{ steps.meta.outputs.version }}
+      - name: Smoke Test
+        run: |
+          docker run --rm -t ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador:${{ steps.meta.outputs.version }} limitador-server --help
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -89,7 +89,8 @@ jobs:
           images: |
             ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador
           tags: |
-            # type=raw,value=${{ github.sha }}
+            # SHA tag for main branch
+            type=raw,value=${{ github.sha }},enable=${{ github.ref_name == env.MAIN_BRANCH_NAME }}
             # set latest tag for main branch
             type=raw,value=latest,enable=${{ github.ref_name == env.MAIN_BRANCH_NAME }}
             # set ref name tag for non-main branches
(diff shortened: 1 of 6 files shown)


limitador-server/Cargo.toml
```diff
@@ -37,6 +37,8 @@ const_format = "0.2.31"
 lazy_static = "1.4.0"
 clap = "4.3"
 sysinfo = "0.29.7"
+openssl = { version = "0.10.57", features = ["vendored"] }
```
KevFan, line 40: - Required updating openssl - https://github.com/Kuadrant/limitador/actions/runs/6573992478/job/17858112986
    - Otherwise will hit this error - https://stackoverflow.com/questions/68871193/pkg-config-error-during-rust-cross-compilation

Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install \
  && dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install epel-release \
  && dnf config-manager --set-enabled powertools
 
-RUN PKGS="gcc-c++ gcc-toolset-12-binutils-gold openssl-devel protobuf-c protobuf-devel git clang kernel-headers" \
+RUN PKGS="gcc-c++ gcc-toolset-12-binutils-gold openssl-devel protobuf-c protobuf-devel git clang kernel-headers perl-IPC-Cmd" \
```
KevFan, line 17: This dockerfile needs to include `perl-IPC-Cmd` to build correctly
- https://github.com/Kuadrant/limitador/actions/runs/6574117673/job/17858507353

Dockerfile.aarch64
KevFan, line 1: Failed to use ubi as the build / run stage due issues mentioned by @eguzki in https://github.com/Kuadrant/limitador/pull/214 https://github.com/Kuadrant/limitador/pull/215

So the main discussion here is are we fine with using community / non-productised images for this dockerfile / build for arm64 ? 🤔 

Personally I dont think it should block this due to:
-  Downstream requirements should generally not leak into the upstream 
- Productised builds should be built on a different infra, which should instead use a "productised" Dockerfile which must use productised images as a requirement.
   -  The main `Dockerfile` which use the ubi image can serve as the "productised" dockerfile which both the amd64 & arm64 image can be built from anyway. Not sure is this "productised" dockerfile usually hosted / provided elsewhere anyway in which case the main dockerfile should follow a similar structure to this dockerfile instead as it's simpler overall
- The quicker build time (20mins) outweighs the downstream requirement to use productised images here since it delays the development flow / community release flow significantly (4 hours to build arm64 image)

Disadvantages to this is that:
- Need to maintain 2 dockerfiles
- The dockerfiles uses different build / runtime stages to each other currently
- Potentially get rate limited by dockerhub occassionally since we use community images

Interested in peoples thoughts on this espeically @alexsnaps @didierofrivia @eguzki @gsaslis

gsaslis: in case it helps, the community `Dockerfile` _is_ different from the product one - but it makes perfect sense to keep the differences to a minimum.

guicassolato: UBI images are publicly available and should be preferred whenever possible. Check out _Reasons to Use UBI_ at https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image.

Personally, I don't see this as leaking requirements from the downstream into the upstream, but as recognizing that behind open source there's a business. We should not kill the business, neither punish the upstream. Instead, we need to find a balance that lets both to thrive as efficient as possible. Generally, UBI is a good compromise.

In contrast, our experience in the past of keeping multiple Dockerfiles and especially with different base images between upstream and downstream has been total chaos. A lot worse than waiting 4 hours for a build to finish. Believe me!

If building for arm64 on UBI base is as inefficient as x12 slower, pressing Red Hat to fix that should be our top priority.

Meanwhile, maybe building for arm64 could be triggered automatically only when merging to `main`/`release-*` branches and manually on-demand?

KevFan: Thanks @guicassolato,  good points ! 👍 I guess the only thing is Red Hat support of rust seems pretty minimal ( https://catalog.redhat.com/software/containers/rhel8/rust-toolset/5b9c6aa85a134643ef2ecc63?architecture=amd64&image=607765c8d70cc51ae2772196 / https://catalog.redhat.com/software/containers/search?gs&q=rust) and the main issue I believe to building on ubi using cross compile is it's missing a `libc6-dev-arm64-cross` equivalent that is available in Debian (https://github.com/Kuadrant/limitador/pull/215). I can reach out to the ubi team on their thoughts to supporting this, but I'm guessing this may take a while.  

I guess my perspective is that for the upstream, I'm not sure do we **need to / should** confine ourselves to using ubi8 at the expense of the developer experience / release flow for limitador itself. Especially if / when the productised build comes, it will be built elsewhere with another dockerfile that needs to be maintained regardless.

Many other Red Hat projects looks to use the community images in their upstream (golang equivalents seems a lot simpler):
- https://github.com/3scale/3scale-operator/blob/master/Dockerfile#L2
- https://github.com/3scale-ops/marin3r/blob/main/Dockerfile#L2

A compromise is that we can have the runtime stage to be ubi but the build stage to use a community image which looks to have worked for me after some tinkering (`docker pull quay.io/kuadrant/limitador:ubi9-test`) 🤔 :
- https://github.com/Kuadrant/limitador/commit/82e978e2670969ae68196e8e49af7f9f0564cac9
- https://github.com/Kuadrant/limitador/actions/runs/6639578417
- https://quay.io/repository/kuadrant/limitador?tab=tags&tag=ubi9-test

Yeah, I agree maintaining multiple Dockerfiles may not be worth it in the end (for example, with the current approach, the arm64 build might break on runtime due to `glibc` versioning mismatch that isn't caught until someone runs this dockerfile / pulls the arm64 image and runs it 🤔). I'll defer to @alexsnaps @didierofrivia @eguzki for the best course of action as they are the main recent contributiors for whether these changes are worth the potential future headache 🤔

gsaslis: To add one more point of view: 

In general, keeping the same parent image between community + product build (and trying to minimize other differences) actually helps us deliver more value **to the community** - because we need to spend less time porting to the product build those community changes that we want to ship as part of the product. That way, we can spend more of our time working on the community project ;)

.github/workflows/build-image.yaml
```diff
@@ -17,43 +16,96 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador
+          tags: |
+            # type=raw,value=${{ github.sha }}
```
KevFan, line 90: Disabled creating per github sha for now so that a unique tag is not created each time while testing changes

.github/workflows/build-image.yaml
```diff
@@ -17,43 +16,96 @@ jobs:
...
+        uses: docker/login-action@v2
         with:
-          image: ${{ steps.build-image.outputs.image }}
-          tags: ${{ steps.build-image.outputs.tags }}
-          registry: ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}
           username: ${{ secrets.IMG_REGISTRY_USERNAME }}
           password: ${{ secrets.IMG_REGISTRY_TOKEN }}
-      - name: Print Image URL
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Create manifest list and push
```
eguzki, line 101: where is the push happening?

Dockerfile.aarch64
eguzki: My 5 cents:

We use `rust` based images in community. It makes life easier (as of today) for the upstream project. And the upstream is what gets the daily work. Downstream is automated and who cares if it takes 20min or 2H? And maybe downstream does not even require cross compiling. At least for ARM. Maybe requires images for RH products in IBM (powerpc and z mainframes).

I reckon the benefits of being aligned (regarding the base image) with downstream and the use of UBI images. But only when the cost is not higher than the benefit. IMO the cost is higher than the benefits as of today.

eguzki: Let me support my opinion with one illustrating example:

* Rust based dockerfile
```dockerfile
FROM rust:1.72 as limitador-build
RUN apt update && apt upgrade -y
RUN apt install -y protobuf-compiler clang
WORKDIR /usr/src/limitador
ARG GITHUB_SHA
ENV GITHUB_SHA=${GITHUB_SHA:-unknown}
ENV RUSTFLAGS="-C target-feature=-crt-static"
COPY . .
RUN cargo build --release``
```
* Ubi based dockerfile
```dockerfile
FROM registry.access.redhat.com/ubi8/ubi:8.7 as limitador-build
ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
ARG RUSTC_VERSION=1.72.0
# the powertools repo is required for protobuf-c and protobuf-devel
RUN dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install \
      http://mirror.centos.org/centos/8-stream/BaseOS/`arch`/os/Packages/centos-gpg-keys-8-6.el8.noarch.rpm \
      http://mirror.centos.org/centos/8-stream/BaseOS/`arch`/os/Packages/centos-stream-repos-8-6.el8.noarch.rpm \
 && dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install epel-release \
 && dnf config-manager --set-enabled powertools
RUN PKGS="gcc-c++ gcc-toolset-12-binutils-gold openssl-devel protobuf-c protobuf-devel git clang kernel-headers" \
    && dnf install --nodocs --assumeyes $PKGS \
    && rpm --verify --nogroup --nouser $PKGS \
    && yum -y clean all
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path --profile minimal --default-toolchain ${RUSTC_VERSION} -c rustfmt -y
WORKDIR /usr/src/limitador
ARG GITHUB_SHA
ENV GITHUB_SHA=${GITHUB_SHA:-unknown}
ENV RUSTFLAGS="-C target-feature=-crt-static"
COPY . .
RUN source $HOME/.cargo/env \
    && cargo build --release
```
UBI based image requires the usage of centos repos in the limitador project (`protobuf-c` and `protobuf-devel` packages)

.github/workflows/build-image.yaml
```diff
@@ -17,43 +16,96 @@ jobs:
...
-          tags: ${{ env.IMG_TAGS }}
-          platforms: linux/amd64,linux/arm64
-          dockerfiles: |
-            ./Dockerfile
+          context: .
+          labels: ${{ steps.meta.outputs.labels }}
           build-args: |
             GITHUB_SHA=${{ github.sha }}
-      - name: Smoke Test
+          cache-from: type=gha,scope=${{ matrix.platform == 'linux/amd64' && 'build-amd' || 'build-arm'}}
```
eguzki, line 52: why not add a `scope` field to the matrix instead of having that shell logic in there?

.github/workflows/build-image.yaml
eguzki: I think we do not need to create image per github SHA. QE admitted they were not using it. Only branch, tags and latest

eguzki:
LGTM

I would merge this and open discussion about the community base image.

Dockerfile.aarch64
eguzki: In the effort to use the UBI as base image, for ARM images, this PR is building on `rust` based images and then in the next stage fitting the binary into an ubi image. This is, IMO, an additional cost of building in UBI for the limitador project. Some extra complexity we will need to maintain and that will be the source of future issues for sure.

.github/workflows/build-image.yaml
pehala: We do use SHA, it is not needed on everything, but at least one sha method of building should be available (e.g. daily builds or pushes to main)

Dockerfile.aarch64
```diff
@@ -0,0 +1,54 @@
+# ------------------------------------------------------------------------------
+# Build Stage cross compiling
+# ------------------------------------------------------------------------------
+
+# Use bullseye as build image instead of Bookworm as ubi9 does not not have GLIBCXX_3.4.30
```
eguzki, line 5: nice!

.github/workflows/build-image.yaml
KevFan: Sure, generating SHA tag only for main branch sounds reasonable 👍

Dockerfile
```diff
@@ -2,19 +2,19 @@
 # Build Stage
 # ------------------------------------------------------------------------------
 
-FROM registry.access.redhat.com/ubi8/ubi:8.7 as limitador-build
+FROM registry.access.redhat.com/ubi9/ubi:9.2 as limitador-build
```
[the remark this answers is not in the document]
KevFan: I can, as long we are "happy" that this Dockefile will no longer be a reflection of what the productised Dockerfile will be 🤔

KevFan:
Rebased with main

.github/workflows/build-image.yaml
```diff
@@ -17,43 +16,98 @@ jobs:
...
+        id: build
+        uses: docker/build-push-action@v5
         with:
-          image: limitador
-          tags: ${{ env.IMG_TAGS }}
-          platforms: linux/amd64,linux/arm64
-          dockerfiles: |
-            ./Dockerfile
+          context: .
+          labels: ${{ steps.meta.outputs.labels }}
```
didierofrivia, line 51: Which labels are set by default from the output of the meta step?

KevFan: This is the JSON output for this build step (taken from the workflow logs):

```json
{
    "tags": [
      "quay.io/kuadrant/limitador:quicker-workflow"
    ],
    "labels": {
      "org.opencontainers.image.created": "2024-01-08T15:23:11.687Z",
      "org.opencontainers.image.description": "Rate limiter",
      "org.opencontainers.image.licenses": "Apache-2.0",
      "org.opencontainers.image.revision": "a39035119c7899901dfc785e[59](https://github.com/Kuadrant/limitador/actions/runs/7449493828/job/20266248072?pr=222#step:4:68)380d52b9b11986",
      "org.opencontainers.image.source": "https://github.com/Kuadrant/limitador",
      "org.opencontainers.image.title": "limitador",
      "org.opencontainers.image.url": "https://github.com/Kuadrant/limitador",
      "org.opencontainers.image.version": "quicker-workflow"
    },
    "annotations": [
      "manifest:org.opencontainers.image.created=2024-01-08T15:23:11.[68](https://github.com/Kuadrant/limitador/actions/runs/7449493828/job/20266248072?pr=222#step:4:77)7Z",
      "manifest:org.opencontainers.image.description=Rate limiter",
      "manifest:org.opencontainers.image.licenses=Apache-2.0",
      "manifest:org.opencontainers.image.revision=a39035119c7899901dfc785e59380d52b9b11986",
      "manifest:org.opencontainers.image.source=https://github.com/Kuadrant/limitador",
      "manifest:org.opencontainers.image.title=limitador",
      "manifest:org.opencontainers.image.url=https://github.com/Kuadrant/limitador",
      "manifest:org.opencontainers.image.version=quicker-workflow"
    ]
  }
```

alexsnaps (approved):

PR #224, opened 2023-11-09
gRPC server reflection

eguzki:
### What

gRPC server reflection provided by [tonic-reflection](https://crates.io/crates/tonic-reflection/0.10.2)

Opt-in feature to be enabled using `--grpc-reflection-service` command line option. 

### Verification steps

Run `sandbox` dev environment

```bash
cd limitador-server/sandbox
make build
make deploy-in-memory
```
> **Note**: `make deploy-in-memory` runs limitador with the  `--grpc-reflection-service` command line option set.

In a separate shell, inspect the RLS GRPC endpoint. First, let's install [grpcurl](https://github.com/fullstorydev/grpcurl) tool. You need [Go SDK 1.8+](https://golang.org/doc/install) installed.

```bash
make grpcurl
```

Inspect `RateLimitService` GRPC service

```bash
bin/grpcurl -plaintext 127.0.0.1:18081 describe envoy.service.ratelimit.v3.RateLimitService
```
The output should be the following
```bash
envoy.service.ratelimit.v3.RateLimitService is a service:
service RateLimitService {
  // Determine whether rate limiting should take place.
  rpc ShouldRateLimit ( .envoy.service.ratelimit.v3.RateLimitRequest ) returns ( .envoy.service.ratelimit.v3.RateLimitResponse );
}
```
If the reflection API is not available the output looks like

```bash
Failed to resolve symbol "envoy.service.ratelimit.v3.RateLimitService": server does not support the reflection API
```

Make a manual request to the RLS endpoint

```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 envoy.service.ratelimit.v3.RateLimitService.ShouldRateLimit <<EOM
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```
> Note that `grpcurl` does not need proto files or compiled protoset files. The reflection service tells `grpc` how to parse request and response.

The response should be the following:

```json
{
  "overallCode": "OK"
}
```

Let's check that rate limiting is working. The limits file looks like

```yaml
- namespace: test_namespace
  max_value: 5
  seconds: 60
  conditions:
    - "req.method == 'POST'"
  variables: []
```

Thus, the limit is set to max 5 request for 60 seconds. Do repeated requests. you should see `OVER_LIMIT` response after 5 requests.

```bash
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```
The expected output

```json
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
```

### DIFF
--- a/limitador-server/src/envoy_rls/server.rs
+++ b/limitador-server/src/envoy_rls/server.rs
@@ -194,6 +194,10 @@ pub fn to_response_header(
     headers
 }
 
+mod rls_proto {
+    pub(crate) const RLS_DESCRIPTOR_SET: &[u8] = tonic::include_file_descriptor_set!("rls");
+}
+
 pub async fn run_envoy_rls_server(
     address: String,
     limiter: Arc<Limiter>,
@@ -202,8 +206,14 @@ pub async fn run_envoy_rls_server(
     let rate_limiter = MyRateLimiter::new(limiter, rate_limit_headers);
     let svc = RateLimitServiceServer::new(rate_limiter);
 
+    let reflection_service = tonic_reflection::server::Builder::configure()
+        .register_encoded_file_descriptor_set(rls_proto::RLS_DESCRIPTOR_SET)
+        .build()
+        .unwrap();
+
     Server::builder()
         .add_service(svc)
+        .add_service(reflection_service)
         .serve(address.parse().unwrap())
         .await
 }
--- a/limitador-server/src/envoy_rls/server.rs
+++ b/limitador-server/src/envoy_rls/server.rs
@@ -202,18 +202,24 @@ pub async fn run_envoy_rls_server(
     address: String,
     limiter: Arc<Limiter>,
     rate_limit_headers: RateLimitHeaders,
+    grpc_reflection_service: bool,
 ) -> Result<(), transport::Error> {
     let rate_limiter = MyRateLimiter::new(limiter, rate_limit_headers);
     let svc = RateLimitServiceServer::new(rate_limiter);
 
-    let reflection_service = tonic_reflection::server::Builder::configure()
-        .register_encoded_file_descriptor_set(rls_proto::RLS_DESCRIPTOR_SET)
-        .build()
-        .unwrap();
+    let reflection_service = match grpc_reflection_service {
+        false => None,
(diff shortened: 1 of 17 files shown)


limitador-server/src/envoy_rls/server.rs
```diff
@@ -202,8 +206,14 @@ pub async fn run_envoy_rls_server(
...
     let svc = RateLimitServiceServer::new(rate_limiter);
 
+    let reflection_service = tonic_reflection::server::Builder::configure()
+        .register_encoded_file_descriptor_set(rls_proto::RLS_DESCRIPTOR_SET)
+        .build()
+        .unwrap();
+
     Server::builder()
         .add_service(svc)
+        .add_service(reflection_service)
```
alexsnaps, line 216: Thinking we may want to add this when _not_ building for target `release`. We can check for [debug_assertions](https://doc.rust-lang.org/stable/reference/conditional-compilation.html#debug_assertions). Or introduce it as a feature? I guess this isn't meant for a production deployment, right?

eguzki: Yes, makes sense. I will give it a try

eguzki: Implemented as command line option, which defaults to not being exposed.

eguzki: The downside of this implementation is that the build will include `tonic-reflection ` crate now. The feature approach could save us doing that. I think it is handier to have a command line option instead of custom compilation to the reflection service working. Let me know what you think,

alexsnaps (approved):

PR #226, opened 2023-11-13
Update rustix

alexsnaps:
Addresses https://github.com/Kuadrant/limitador/security/dependabot/16


sunfishcode:
According to the checked-in Cargo.lock, limitador depends on rustix through is-terminal, tempfile, and which, and none of those use `rustix::fs::Dir`, so it is not affected by GHSA-c827-hfw6-qwvm.

eguzki (approved):

alexsnaps:
@sunfishcode Thanks! Yeah… it's mostly so dependatbot is silenced. Thanks for your work tho 🙏

PR #244, opened 2023-12-14
Sad state of the world

alexsnaps:


alexsnaps:
I wanna cry... all of this is broken beyond belief:

 - For some reason, I can't access the project thru the GraphQL API: `Resource not accessible by integration`
 - Couldn't find any way to enable the reads for the action
 - So fell back to just skipping if the secret isn't there... sad
 - Also know that for some reason, on some conditions, GH just decides to skip this action and I had to create I don't know how many PRs to test this

adam-cattermole (approved):

.github/workflows/issues-workflow.yaml
```diff
@@ -13,7 +13,14 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
+      - uses: svrooij/secret-gate-action@v1
+        id: secret
+        with:
+          inputsToCheck: "ADD_ISSUES_TOKEN"
+        env:
+          ADD_ISSUES_TOKEN: ${{ secrets.ADD_ISSUES_TOKEN }}
       - uses: actions/add-to-project@v0.5.0
+        if: steps.secret.outputs.inputsChecked == 'true'
```
eguzki, line 23: did you try the following?? 

```
if: secrets.ADD_ISSUES_TOKEN != ''
```
Maybe it works (I have not tried myself)

eguzki: You might want to check this https://github.com/Kuadrant/kuadrant-operator/pull/385

Implementing the same for the kuadrant opeator

PR #261, opened 2024-02-20
Support pushing traces to otel collector

adam-cattermole:
This adds support to push traces to an otel collector, configured via new command line options, `--tracing-host` and `--tracing-port`.

Currently this is configured to use grpc without TLS however this could be added as a follow up.

## Verification:

For this example I'm running Redis as the datastore to verify.

### Run Redis service

```sh
docker run --rm -it -p 6379:6379 redis
```

### Run Jaeger with OTLP

```sh
docker run --rm -it --name jaeger -e COLLECTOR_OTLP_ENABLED=true -p 16686:16686 -p 4317:4317 -p 4318:4318 jaegertracing/all-in-one:latest
```

```sh
firefox -private-window "localhost:16686"
```

### Build and run from this branch

```sh
cargo build --bin limitador-server
```

```sh
./target/debug/limitador-server --tracing-endpoint rpc://localhost:4317 -vvvv --grpc-reflection-service limitador-server/sandbox/limits.yaml redis redis://127.0.0.1
```

### Send Requests

```sh
grpcurl -plaintext -d @ 127.0.0.1:8081 envoy.service.ratelimit.v3.RateLimitService.ShouldRateLimit <<EOM
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```

See the spans for the requests under the `limitador-server` service in the jaeger ui:

image

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -31,6 +31,9 @@ use limitador::{
 };
 use notify::event::{ModifyKind, RenameMode};
 use notify::{Error, Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
+use opentelemetry::KeyValue;
+use opentelemetry_otlp::{Protocol, WithExportConfig};
+use opentelemetry_sdk::{trace, Resource};
 use std::env::VarError;
 use std::fs;
 use std::path::Path;
@@ -42,6 +45,8 @@ use thiserror::Error;
 use tokio::runtime::Handle;
 use tracing::level_filters::LevelFilter;
 use tracing_subscriber::fmt::format::FmtSpan;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
 
 mod envoy_rls;
 mod http_api;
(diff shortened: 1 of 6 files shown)


limitador-server/src/main.rs
```diff
@@ -290,12 +295,37 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                ])))
+                .install_batch(opentelemetry_sdk::runtime::Tokio)?;
+            let telemetry_layer = tracing_opentelemetry::layer().with_tracer(tracer);
+            tracing_subscriber::registry()
+                .with(level)
+                .with(fmt_layer)
+                .with(telemetry_layer)
+                .init();
+        } else {
+            tracing_subscriber::registry()
```
eguzki, line 324: what is this doing? (when exporter is not being defined)

adam-cattermole: The tracing package is being used for the standard logging since #250 - if `tracing_host` is unset we ensure that the `fmt_layer` is still applied so we get log output on stdout

limitador-server/src/main.rs
```diff
@@ -290,12 +295,37 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        };
+
+        if !config.tracing_host.is_empty() {
+            let tracer = opentelemetry_otlp::new_pipeline()
+                .tracing()
+                .with_exporter(
+                    opentelemetry_otlp::new_exporter()
+                        .tonic()
+                        .with_endpoint(format!("rpc://{}", config.tracing_address()))
+                        .with_protocol(Protocol::Grpc),
```
alexsnaps, line 311: I think I'd want to align that with authorino's way of configuring...

adam-cattermole: I've had a look and this `.with_protocol` doesn't actually effect the behaviour. Either we use `.tonic()` and it's grpc, or we use `.http()` and provide our choice of http client. For now I've moved to a `--tracing-endpoint` option which contains scheme the same as authorino, but it won't use http if provided unless we check for the provided scheme and configure an http client

alexsnaps (approved):
LGTM, modulo the unnecessary diffs, but I'll let you be the judge of their faith

PR #267, opened 2024-03-12
New time metric aggregation layer and refactor of prometheus metrics

adam-cattermole:
Built on top of #261 which should be merged first

### Adds a new metrics layer to the tracing configuration which times accesses to different spans.

`gather` accepts:
- `group`: the parent span to combine for
- `consumer`: a function to perform some kind of operation on the `Timings` collected for this `group`
- `records`: a list of different sub-spans of `group` that should be summed together

### Refactor

- Moved prometheus_metrics out of the library and now defined at the server level


eguzki:
I am trying to understand what is this doing. 
I tried this branch and compared with`main`. Same data at `/metrics` and same info from opentelemetry spans I could see in jaeger UI.

adam-cattermole:
The primary difference is the way that those metrics are collected at /metrics @eguzki. Before the calculation was all being done inline with time instants wrapping the sections of code that access the datastore in the library. It's now been abstracted outside of the library, with all of the latency metrics being calculated through a new metrics layer instrumenting the spans. The result is two primary changes:

1. The library no longer cares or implements metrics and moves the owner of the metrics to limitador-server
2. A new tracing metrics layer that times and sums spans based on configuration, and aggregates based on the provided consumer function

limitador-server/src/metrics.rs
```diff
@@ -0,0 +1,291 @@
...
+            busy: 0,
+            last: Instant::now(),
+        }
+    }
+}
+
+impl ops::Add for Timings {
+    type Output = Self;
+
+    fn add(self, _rhs: Self) -> Self {
```
alexsnaps, line 29: It is "unusual" to prefix a variable with an underscore, when you actually use it... any reason?

limitador-server/src/metrics.rs
```diff
@@ -0,0 +1,291 @@
...
+        Self {
+            busy: self.busy + _rhs.busy,
+            idle: self.idle + _rhs.idle,
+            last: if self.last < _rhs.last {
```
alexsnaps, line 33: See [`max`](https://doc.rust-lang.org/std/cmp/trait.Ord.html#method.max) which is implemented for `Instant` if I'm not mistaken

limitador-server/src/metrics.rs
```diff
@@ -0,0 +1,291 @@
...
+                _rhs.last
+            } else {
+                self.last
+            },
+        }
+    }
+}
+
+impl ops::AddAssign for Timings {
+    fn add_assign(&mut self, _rhs: Self) {
```
alexsnaps, line 43: `rhs` instead?

limitador-server/src/metrics.rs
```diff
@@ -0,0 +1,291 @@
...
+            } else {
+                // otherwise create a new SpanState with ourselves
+                extensions.insert(SpanState::new(name.to_string()))
+            }
+        }
+
+        if let Some(span_state) = extensions.get_mut::<SpanState>() {
+            // either we are an aggregator or nested within one
+            for group in span_state.group_times.keys() {
+                for record in &self.groups.get(group).unwrap().records {
```
alexsnaps, line 143: Could you use `expect` instead of `unwrap` here... just so avoid adding yet more of them to the code base until we address #280

alexsnaps (approved):
Minor stuff... but otherwise LGTM

limitador-server/src/metrics.rs
adam-cattermole: Yep good question that is strange, I just used the variable that my IDE auto-generated for me but I guess it was because I wasn't using it yet - I'll change

PR #276, opened 2024-03-21
Write behind lock

### LINKED ISSUE #271: Fix locking in the `Batcher`
Currently, the `Mutex` that protects the pending writes is held for the duration of doing these (individual) writes to Redis… So that all writes queue up, waiting. Ideally, the writes should be truly asynchronous and not block the in-band path.

alexsnaps:
Fix #271


eguzki (approved):
LGTM

Tested with https://github.com/Kuadrant/architecture/issues/87 and https://github.com/Kuadrant/architecture/issues/88

PR #279, opened 2024-03-22
Resilient cache

### LINKED ISSUE #273: Make `CachedRedisStorage` fault tolerant
In the vain of #268, the `CachedRedisStorage` should incorporate the resiliency and use its cache as the fallback datastore.

alexsnaps:
builds on #276 
fixes #273


eguzki:
TBH this is very hard to follow. I can only say that I am happy to test this whenever it is ready to test.

alexsnaps:
> TBH this is very hard to follow. I can only say that I am happy to test this whenever it is ready to test.

I guess this is the important change here: cb98725260da82efbbb857d24a7906f9b74cbd24
Which builds upon this change the `StorageErr` that now exposes transient errors as such... mostly the ones triggered because of a timeout awaiting a response from Redis, controlled thru this change: 2f5e6cd9b3a4f915a318660f9d2c25195e899437

If you want to test this, you can use the limitador server built from this branch using: 
```
redis_cached redis://127.0.0.1:6379 --ratio 1 --response-timeout 1
```

This will require your redis to always respond in at most 1ms. You can also shut the redis down entirely... I use [my limitador-driver](https://github.com/alexsnaps/limitador-driver) to put load on Limitador during these tests and... well you shouldn't notice any difference and _always_ observe single digit millisecond latency to Limitador from the driver.

alexsnaps:
@eguzki if you have that integration test for `CachedRedisStorage`, running it on top of this branch _should_ make it pass (until proven wrong!)

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -93,7 +94,7 @@ impl AsyncCounterStorage for CachedRedisStorage {
                             first_limited = Some(a);
                         }
                         if load_counters {
-                            counter.set_remaining(val);
+                            counter.set_remaining(counter.max_value() - val);
```
didierofrivia, line 97: Maybe we could get rid of this, no? Since we are unifying the way we apply limits, we won't keep track of remaining because of subtracting but checking if we reach the limit instead

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -210,30 +226,56 @@ impl CachedRedisStorage {
...
+        response_timeout: Duration,
     ) -> Result<Self, RedisError> {
         let info = ConnectionInfo::from_str(redis_url)?;
-        let redis_conn_manager = ConnectionManager::new(
+        let redis_conn_manager = ConnectionManager::new_with_backoff_and_timeouts(
             redis::Client::open(info)
                 .expect("This couldn't fail in the past, yet now it did somehow!"),
+            2,
+            100,
+            6,
```
didierofrivia, line 237: Maybe it could be more informative if we extract this to default values in a `const` (?)

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -155,21 +168,23 @@ impl AsyncCounterStorage for CachedRedisStorage {
...
+                Self::batch_counter(delta, &mut batcher, counter);
             }
         } else {
             for counter in counters.iter() {
-                self.update_counter(counter, delta).await?
+                self.update_counter(counter, delta).await.or_else(|err| {
+                    if err.is_transient() {
+                        self.partitioned(true);
+                        let mut batcher = self.batcher_counter_updates.lock().unwrap();
+                        Self::batch_counter(delta, &mut batcher, counter);
```
didierofrivia, line 182: We are just collecting and updating the counters in memory only to this point, right?

limitador/src/storage/redis/redis_cached.rs
alexsnaps: They are... they actually are inlined from the redis crate, cause I can't reuse theirs ... not public.

limitador/src/storage/redis/redis_cached.rs
alexsnaps: Not sure what you mean... that's when they are flushed to redis... yes. Always was the case btw, this isn't much different from before.

didierofrivia: then that batch should be sent to redis once we recover the connection, right? I reckon we don't consider the case the previous redis instance died for good

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -138,7 +138,9 @@ impl AsyncCounterStorage for CachedRedisStorage {
                         counter_ttls_msecs[i],
                         ttl_margin,
                     );
-                    let remaining = counter_vals[i].unwrap_or(counter.max_value()) - delta;
+                    let remaining = counter.max_value()
+                        - counter_vals[i].unwrap_or(counter.max_value())
```
eguzki, line 143: Shouldn't be `0` ??
`counter_vals[i].unwrap_or(0)` ??

Like I set in https://github.com/Kuadrant/limitador/blob/main/limitador/src/storage/redis/redis_async.rs#L123

eguzki:
Looking good codewise. 

Before giving my :+1: , I want to run some tests.

eguzki:
> @eguzki if you have that integration test for `CachedRedisStorage`, running it on top of this branch _should_ make it pass (until proven wrong!)

Unfortunately did not pass. 

```bash
failures:

---- test::check_rate_limited_and_update_load_counters_with_async_redis_and_local_cache stdout ----
thread 'test::check_rate_limited_and_update_load_counters_with_async_redis_and_local_cache' panicked at limitador/tests/integration_tests.rs:698:13:
assertion failed: !result.limited
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::check_rate_limited_and_update_with_async_redis_and_local_cache stdout ----
thread 'test::check_rate_limited_and_update_with_async_redis_and_local_cache' panicked at limitador/tests/integration_tests.rs:657:13:
assertion failed: !rate_limiter.check_rate_limited_and_update(namespace, &values, 1,
                    false).await.unwrap().limited


failures:
    test::check_rate_limited_and_update_load_counters_with_async_redis_and_local_cache
    test::check_rate_limited_and_update_with_async_redis_and_local_cache

test result: FAILED. 196 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.39s

error: test failed, to rerun pass `-p limitador --test integration_tests`
```

I will be looking into this as well

limitador/src/storage/redis/redis_cached.rs
alexsnaps: Yes!

limitador/src/storage/redis/redis_cached.rs
alexsnaps: Which part? Setting that remainder is to set the headers in the response to the user, to inform them of how many request they still can perform in ... (see line below) the time they have left on their counter. 
There isn't much for us to decide, as that's all part of the [draft RFC](https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/)

limitador/src/storage/redis/redis_cached.rs
alexsnaps: So @didierofrivia did spot an issue here, we might be partitioned and _not having batching enabled_ which would result in the batch size growing on every partition without ever flushing to Redis... 

I'm thinking we should _always_ enable batching as: 
 - It optimizes most for the happy path (i.e. limits not reached)
 - We are planning on using this implementation for Global RL, which adds such a big latency on going to redis... 

Thoughts?

alexsnaps:
> Unfortunately did not pass.

That's probably that math I had missed...

limitador/src/storage/redis/redis_cached.rs
didierofrivia: I agree with you that batching should be always enabled 👍🏼

eguzki:
Before running partitioning tests, I run simple test of running 1 request per second hitting a counter with a limit of max 5 req every 10 secs. And results are not the expected ones. 

### Test

* Checkout this code branch and enter sandbox folder

```bash
cd limitador-server/sandbox/
```

* Build limitador image `limitador-testing` for the current branch
```bash
make build
```

* Run sandbox with redis 

The applied configuration is:
* max_ttl: 30secs
* ratio: 1
* flush period: 100ms
```diff
diff --git a/limitador-server/sandbox/docker-compose-limitador-redis-cached.yaml b/limitador-server/sandbox/docker-compose-limitador-redis-cached.yaml
index 6db3a4d..eb56293 100644
--- a/limitador-server/sandbox/docker-compose-limitador-redis-cached.yaml
+++ b/limitador-server/sandbox/docker-compose-limitador-redis-cached.yaml
@@ -21,11 +21,11 @@ services:
       - /opt/kuadrant/limits/limits.yaml
       - redis_cached
       - --ttl
-      - "3"
+      - "30000"
       - --ratio
-      - "11"
+      - "1"
       - --flush-period
-      - "2"
+      - "100"
       - --max-cached
       - "5000"
       - redis://redis:6379
diff --git a/limitador-server/sandbox/limits.yaml b/limitador-server/sandbox/limits.yaml
index 0f7b90b..4d553b0 100644
--- a/limitador-server/sandbox/limits.yaml
+++ b/limitador-server/sandbox/limits.yaml
@@ -1,13 +1,13 @@
 ---
 - namespace: test_namespace
-  max_value: 10
-  seconds: 60
+  max_value: 5
+  seconds: 10
   conditions:
     - "req.method == 'GET'"
   variables: []
 - namespace: test_namespace
   max_value: 5
-  seconds: 60
+  seconds: 10
   conditions:
     - "req.method == 'POST'"
   variables: []
```

```bash
make deploy-redis
```
Limitador reports the following configuration as expected
```
limitador-1  | 2024-03-28T23:23:26.730508Z  INFO limitador_server: Using config: Configuration { limits_file: "/opt/kuadrant/limits/limits.yaml", storage: Redis(RedisStorageConfiguration { url: "redis://redis:6379", cache: Some(RedisStorageCacheConfiguration { flushing_period: 100, max_ttl: 30000, ttl_ratio: 1, max_counters: 5000, response_timeout: 350 }) }), rls_host: "0.0.0.0", rls_port: 8081, http_host: "0.0.0.0", http_port: 8080, limit_name_in_labels: false, tracing_endpoint: "", log_level: Some(LevelFilter::DEBUG), rate_limit_headers: None, grpc_reflection_service: true }
```


* Open a new shell and run 1 request per second

```bash
make grpcurl
```

```bash
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```

I would expect on every 10 seconds, 5 request being accepted and 5 requests being rejected. Instead I see this

```
{
  "overallCode": "OK"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
```

It could be some missing store increment fixes I added in https://github.com/Kuadrant/limitador/pull/283, specifically https://github.com/Kuadrant/limitador/pull/283/files#diff-d91ad49639aa408abab791b66f7d3bca5ea05d4ffff21d6afd7f5034b378774aR159

alexsnaps:
@eguzki take it for another round, this should be good.
If so, lets review/merge this and iterate from there on adding tests & I'll refactor this TTL et al stuff into a `CachedValue` struct

eguzki:
Now looking good
With the limit 
```yaml
- namespace: test_namespace
  max_value: 5
  seconds: 10
  conditions:
    - "req.method == 'POST'"
  variables: []
```
The result is as expected, every 10 secs, 5 accepted and 5 rejected

```
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OVER_LIMIT"
}
{
  "overallCode": "OK"
}
{
  "overallCode": "OK"
}
```

And integration tests for the redis cached datastore passing :ok:

eguzki:
I have run the same test multiple times: 
* while sending 1req/sec, kill redis. See what happens and then bring redis back on. 

What I see is that when redis is down, limitador starts returning "always" `"overallCode": "OK"` and limits are not respected. When redis is back on, the responses are again consistent with the limits set. 

In one of the test (I could not reproduce), limitador entered "poison" state and it was returning "always"
```
ERROR:
  Code: Unavailable
  Message: Service unavailable
```
I do not know what happens when redis is back, because I screwed up the environment when trying to get redis back on, and I could not reproduce the issue.

alexsnaps:
Damn... seconds aren't milliseconds! 🤦 
Should really be fixed now... BUT! You'd still see an "error" when resolving the partition, which is what we actually expected.
i.e. when it reconnects to the Redis, it writes _all_ the increments that happened during the partition, regardless if they still should be or not... Again, that was always broken tho. 
So I'd rather have it fixed as a follow up PR as _someone_ already pointed out:

>TBH this is very hard to follow.

😉

alexsnaps:
also, on the poisoning, it means something went wrong while holding a lock… e.g. an `.unwrap` on an `Err` that had the thread panic, while that thread held the mutex… See #280 , but I'll give it a pass in this code when refactoring the `CachedCounterValue` already

eguzki (approved):
working like a charm. 

when redis is down, requests keep being rate limited according with the configured limits. And when redis is respawned, after some "over rejecting" period (rejecting more requests than expected), the behavior is consistent with the limits. The "over rejecting" must be due to limitador flushing cached counters of past period hits.

PR #281, opened 2024-03-25
store increments in redis

### LINKED ISSUE #269: Store increments in Redis
Currently we store decrements in Redis. We first store the limit's threshold on miss and then decrement 

Current `SCRIPT_UPDATE_COUNTER` script:
```lua
local set_res = redis.call('set', KEYS[1], ARGV[1], 'EX', ARGV[2], 'NX')
redis.call('incrby', KEYS[1], - ARGV[3])
if set_res then
    redis.call('sadd', KEYS[2], KEYS[1])
end
```

The idea is to completely decouple the counter from the limit and only store what happened, i.e. increments from `0`
Something along these lines:

```lua
local c = redis.call('incrby', KEYS[1], ARGV[2])
if c == ARGV[2] then
  redis.call('expire', KEYS[1], ARGV[1], 'NX')
  redis.call('sadd', KEYS[2], KEYS[1])
end
return c
```

Let's introduce this as another script and try leveraging it in our both our `(Async)CounterStorage`s

eguzki:
### What

Fixes #269 

:warning: Breaks backward compatibility

### Verification steps

* Checkout this code branch

```bash
cd limitador-server/sandbox/
```

* Build limitador image `limitador-testing` for the current branch
```bash
make build
```

* Run sandbox with redis 

The applied limit is 
```yaml
- namespace: test_namespace 
  max_value: 10             
  seconds: 60               
  conditions:               
  - "req.method == 'GET'" 
  variables: []             
- namespace: test_namespace 
  max_value: 5              
  seconds: 60               
  conditions:               
  - "req.method == 'POST'"
  variables: [] 
```

```bash
make deploy-redis
```

* Open a new shell

* Let's create a counter running on `ShouldRateLimit` request

```bash
make grpcurl
```

```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 envoy.service.ratelimit.v3.RateLimitService.ShouldRateLimit <<EOM
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```

* Check the counter value
```
docker compose -p sandbox exec redis redis-cli get "namespace:{test_namespace},,counter:{\"limit\":{\"namespace\":\"test_namespace\",\"seconds\":60,\"conditions\":[\"req.method == \\\"POST\\\"\"],\"variables\":[]},\"set_variables\":{},\"remaining\":null,\"expires_in\":null}"
```
> Counters have TTL of 60sec, so do not go for a coffee between the request and the redis counter read.

That should be `1`

* Run yet another `ShouldRateLimit` request
```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 envoy.service.ratelimit.v3.RateLimitService.ShouldRateLimit <<EOM
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```

* Check the counter value
```bash
docker compose -p sandbox exec redis redis-cli get "namespace:{test_namespace},,counter:{\"limit\":{\"namespace\":\"test_namespace\",\"seconds\":60,\"conditions\":[\"req.method == \\\"POST\\\"\"],\"variables\":[]},\"set_variables\":{},\"remaining\":null,\"expires_in\":null}"
```

That should be `2`

* run N times

```bash
{
    "domain": "test_namespace",
    "hits_addend": 1,
    "descriptors": [
        {
            "entries": [
                {
                    "key": "req.method",
                    "value": "POST"
                }
            ]
        }
    ]
}
EOM
```

After few hits, the response should be `"overallCode": "OVER_LIMIT"` and the counter value should be `5`

```bash
docker compose -p sandbox exec redis redis-cli get "namespace:{test_namespace},,counter:{\"limit\":{\"namespace\":\"test_namespace\",\"seconds\":60,\"conditions\":[\"req.method == \\\"POST\\\"\"],\"variables\":[]},\"set_variables\":{},\"remaining\":null,\"expires_in\":null}"
```

### Tear down sandbox

```bash
make clean
```

### DIFF
--- a/limitador/src/storage/redis/mod.rs
+++ b/limitador/src/storage/redis/mod.rs
@@ -47,7 +47,8 @@ pub fn is_limited(
 
     let mut first_limited = None;
     for (i, counter) in counters.iter_mut().enumerate() {
-        let remaining = counter_vals[i].unwrap_or(counter.max_value()) - delta;
+        // remaining  = max - (curr_val + delta)
+        let remaining = counter.max_value() - (counter_vals[i].unwrap_or(0) + delta);
         counter.set_remaining(remaining);
         let expires_in = counter_ttls_msecs[i]
             .map(|x| {
--- a/limitador/src/storage/redis/scripts.rs
+++ b/limitador/src/storage/redis/scripts.rs
@@ -9,15 +9,15 @@
 
 // KEYS[1]: counter key
 // KEYS[2]: key that contains the counters that belong to the limit
-// ARGV[1]: counter max val
-// ARGV[2]: counter TTL
-// ARGV[3]: delta
+// ARGV[1]: counter TTL
+// ARGV[2]: delta
 pub const SCRIPT_UPDATE_COUNTER: &str = "
-    local set_res = redis.call('set', KEYS[1], ARGV[1], 'EX', ARGV[2], 'NX')
-    redis.call('incrby', KEYS[1], - ARGV[3])
-    if set_res then
-        redis.call('sadd', KEYS[2], KEYS[1])
-    end";
+    local c = redis.call('incrby', KEYS[1], ARGV[2])
+    if c == tonumber(ARGV[2]) then
+      redis.call('expire', KEYS[1], ARGV[1], 'NX')
+      redis.call('sadd', KEYS[2], KEYS[1])
+    end
+    return c";
 
 // KEYS: the function returns the value and TTL (in ms) for these keys
 // The first position of the list returned contains the value of KEYS[1], the
--- a/limitador/src/storage/redis/scripts.rs
+++ b/limitador/src/storage/redis/scripts.rs
@@ -14,7 +14,7 @@
 pub const SCRIPT_UPDATE_COUNTER: &str = "
     local c = redis.call('incrby', KEYS[1], ARGV[2])
     if c == tonumber(ARGV[2]) then
-      redis.call('expire', KEYS[1], ARGV[1], 'NX')
+      redis.call('expire', KEYS[1], ARGV[1])
       redis.call('sadd', KEYS[2], KEYS[1])
     end
     return c";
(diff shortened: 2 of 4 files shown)


limitador/src/storage/redis/scripts.rs
```diff
@@ -14,7 +14,7 @@
 pub const SCRIPT_UPDATE_COUNTER: &str = "
     local c = redis.call('incrby', KEYS[1], ARGV[2])
     if c == tonumber(ARGV[2]) then
-      redis.call('expire', KEYS[1], ARGV[1], 'NX')
+      redis.call('expire', KEYS[1], ARGV[1])
```
eguzki, line 17: According to redis doc https://redis.io/commands/expire/
```
History

    Starting with Redis version 7.0.0: Added options: NX, XX, GT and LT.

```

So, removing the `NX` option to keep compatibility with redis 5 and redi 6.

Actually, I think it is not needed, as the key has just been created, so expiration has not been set. This happens within the context of the script, so no race condition issues can hit here.

alexsnaps: Strictly speaking we indeed don't need it...

limitador/src/storage/redis/mod.rs
```diff
@@ -47,7 +47,8 @@ pub fn is_limited(
 
     let mut first_limited = None;
     for (i, counter) in counters.iter_mut().enumerate() {
-        let remaining = counter_vals[i].unwrap_or(counter.max_value()) - delta;
+        // remaining  = max - (curr_val + delta)
+        let remaining = counter.max_value() - (counter_vals[i].unwrap_or(0) + delta);
```
alexsnaps, line 51: Not for this PR, but I'm thinking of folding most of that logic within `Counter` directly soon... probably along side other methods actually.

alexsnaps (approved):

PR #285, opened 2024-03-28
Batch update

didierofrivia:
This PR replaces the way Limitador, configured to use _Redis Cached_ storage, updates (flushes the in memory cache) to Redis. Instead of updating counters one by one, Limitador sends the batch of counters to update to Redis.

### TODO

* [x] Batch updating to redis
* [x] Fix rebase conficts
* [ ] ~~Limit batch update counters to redis (?)~~ for the near future
* [x] Get back the values from redis...
* [x]  and update cache (?)
* [x] Test that update counters are being updated in redis
* [x] Test that the cached counters are updated

### Notes

* In a setup where the Redis instance is (physically) far from the Limitador instance, it could happen that we encounter network partition error when the _response timeout_ is exceeded:
   #### EC Frankfurt – Limitador Sao Paulo
```sh
Executing limitador: $ limitador-server /bench/limits.yaml redis_cached redis://test-limitador-redis.jeigrh.ng.0001.euc1.cache.amazonaws.com:6379
Limitador Server v1.4.0-dev (b7e4d87c) [] release
Executing limitador-driver: $ limitador-driver rpc://127.0.0.1:8081 1m
Hitting Limitador at rpc://127.0.0.1:8081 for a total duration of 1 minutes
Current: hits: 12195, mean: 0.081ms, p50: 0.062ms, p90: 0.073ms, p99: 0.092ms, p999: 0.106ms, max: 201.727ms
Current: hits: 15461, mean: 0.064ms, p50: 0.062ms, p90: 0.074ms, p99: 0.091ms, p999: 0.112ms, max: 0.335ms
Current: hits: 15367, mean: 0.064ms, p50: 0.062ms, p90: 0.074ms, p99: 0.092ms, p999: 0.112ms, max: 0.399ms
2024-04-03T14:58:00.941325Z ERROR limitador::storage::redis::redis_cached: Partition to Redis detected!
Current: hits: 6912, mean: 0.144ms, p50: 0.062ms, p90: 0.074ms, p99: 0.092ms, p999: 0.137ms, max: 352.255ms
Current: hits: 15414, mean: 0.064ms, p50: 0.062ms, p90: 0.074ms, p99: 0.091ms, p999: 0.107ms, max: 0.363ms
2024-04-03T14:58:02.941271Z ERROR limitador::storage::redis::redis_cached: Partition to Redis detected!
```

### DIFF
--- a/limitador/src/storage/redis/redis_cached.rs
+++ b/limitador/src/storage/redis/redis_cached.rs
@@ -19,7 +19,7 @@ use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant, SystemTime};
-use tracing::{error, trace_span, warn, Instrument};
+use tracing::{debug_span, error, warn, Instrument};
 
 // This is just a first version.
 //
@@ -233,9 +233,7 @@ impl CachedRedisStorage {
             .ttl_ratio_cached_counter(ttl_ratio_cached_counters)
             .build();
 
-        let cacher = Arc::new(cached_counters);
-        let cacher_clone = cacher.clone();
-
+        let counters_cache = Arc::new(cached_counters);
         let partitioned = Arc::new(AtomicBool::new(false));
         let async_redis_storage =
             AsyncRedisStorage::new_with_conn_manager(redis_conn_manager.clone());
@@ -244,6 +242,7 @@ impl CachedRedisStorage {
 
         {
             let storage = async_redis_storage.clone();
+            let counters_cache_clone = counters_cache.clone();
             let conn = redis_conn_manager.clone();
             let p = Arc::clone(&partitioned);
             let batcher_flusher = batcher.clone();
@@ -254,7 +253,7 @@ impl CachedRedisStorage {
                         conn.clone(),
                         batcher_flusher.clone(),
                         storage.is_alive(),
-                        cacher_clone.clone(),
+                        counters_cache_clone.clone(),
                         p.clone(),
                     )
                     .await;
@@ -264,7 +263,7 @@ impl CachedRedisStorage {
         }
 
         Ok(Self {
-            cached_counters: cacher,
+            cached_counters: counters_cache,
             batcher_counter_updates: batcher,
             redis_conn_manager,
             async_redis_storage,
@@ -409,7 +408,7 @@ async fn update_counters<C: ConnectionLike>(
         }
     }
 
-    let span = trace_span!("datastore");
+    let span = debug_span!("datastore");
     // The redis crate is not working with tables, thus the response will be a Vec of counter values
     let script_res: Vec<i64> = script_invocation
         .invoke_async(redis_conn)


eguzki:
you should rebase your branch to get some fixes in redis_cached

limitador/src/storage/redis/redis_async.rs
```diff
@@ -231,6 +233,30 @@ impl AsyncRedisStorage {
 
         Ok(())
     }
+
+    pub async fn update_counters(
```
didierofrivia, line 237: this fn could be named differently, a more explicit name would be `batch_update_counters`... will leave it to the review

limitador/Cargo.toml
```diff
@@ -54,6 +54,7 @@ base64 = { version = "0.22", optional = true }
 [dev-dependencies]
 serial_test = "3.0"
 criterion = { version = "0.5.1", features = ["html_reports"] }
+redis-test = { version = "0.4.0", features = ["aio"] }
```
alexsnaps, line 57: This broke the wasm build. It is really a dev-dependency (only for tests), so moved it here. 
If you need to declare a dependency that's only for a feature tho, but a runtime dep - like what you did here. There are two things you need to do: 
 - Mark it as `option = true`
 - Add it to the list of dependencies for that feature 

```toml
 We make redis and infinispan optional to be able to compile for wasm32.
[features]
default = ["disk_storage", "redis_storage"]
disk_storage = ["rocksdb"]
redis_storage = ["redis", "r2d2", "tokio"]
infinispan_storage = ["infinispan", "reqwest", "base64", "tokio"]
lenient_conditions = []
```

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -226,52 +227,44 @@ impl CachedRedisStorage {
         )
         .await?;
 
+        let cached_counters = CountersCacheBuilder::new()
+            .max_cached_counters(max_cached_counters)
+            .max_ttl_cached_counter(ttl_cached_counters)
+            .ttl_ratio_cached_counter(ttl_ratio_cached_counters)
+            .build();
+
+        let cacher = Arc::new(cached_counters);
```
alexsnaps, line 236: `cacher`? Why the rename? what does `cacher` mean?

alexsnaps (approved):
This looks good to me… other that rename of `cached_counters` into `cacher` that I don't understand (like I don't understand the binding's name… literally). 
Now, I also think the changes to `keys.rs` of removing the `namespace: &str` from the method signature are… unfortunate. As fundamentally I think we shouldn't resend the keys back, as we could "just" send the value & ttl for each ordered as we passed them in. This would avoid that change, but mostly would save bandwidth… We can iterate over these changes tho. Again, in the vein of avoiding doing work we can avoid doing.

limitador/src/storage/redis/redis_cached.rs
didierofrivia: I reckon naming it that way because of the similar composition with the "batcher", but it's true that in this case it doesn't reflect well what it is. I could rename it to `counters_cache`

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -395,10 +388,108 @@ impl CachedRedisStorageBuilder {
...
+            script_invocation.key(key_for_counter(&counter));
+            script_invocation.key(key_for_counters_of_limit(counter.limit()));
+            script_invocation.arg(counter.seconds());
+            script_invocation.arg(delta);
+            // We need to store the counter in the actual order we are sending it to the script
+            res.push((counter, 0, 0));
+        }
+    }
+
+    let span = trace_span!("datastore");
```
adam-cattermole, line 412: I wonder if it would make sense to have this a `debug_span` to match the other `datastore` spans - I don't think it particularly matters though

adam-cattermole (approved):
I think these changes look good to me, tested it out with the driver and can see the writes in redis

PR #288, opened 2024-04-05
Cached value

alexsnaps:
This fixes the issue when a partition lasts longer than a counter's time window and all the increments from the previous window get batched with the current one, resulting in limiting the requests too early


alexsnaps (commented):
As you can see I'm not entirely happy with things as they are... 
But this could be a start, and certainly fixes the issue we had. In any case, putting this up for review by you smart folks! I need some input...

didierofrivia (approved):
Makes more sense to use moka's cache and keep track of the counter values with the `AtomicExpiringValue`, also, a nice refactor done to that one. This changes drastically the way I'll be updating the cache once I get the updated values from Redis, so hopefully will be merged soon :)

adam-cattermole (approved):
This also looks good to me. The changes all seem to be make sense and it's nice to have all the logic contained in these new types

PR #290, opened 2024-04-16
Update almost all deps

alexsnaps:


limitador/Cargo.toml
```diff
@@ -49,10 +49,10 @@ tokio = { version = "1", optional = true, features = [
 ] }
 infinispan = { version = "0.3", optional = true }
 reqwest = { version = "0.11", optional = true }
```
alexsnaps, line 51: This should be `0.12` but `infinispan` uses `0.11` and has `reqwest` types as part of its API, so the two are bound

adam-cattermole (approved):

PR #291, opened 2024-04-18
Refactor the way we deal with partitions

alexsnaps:
This also makes for more consistent logging of the encountering a partition... 
There is a todo that we need to fix, _not_ losing the counters that have not been written out to Redis, when the partition occurs in the middle of flushing the batch.


limitador/src/storage/redis/redis_cached.rs
```diff
@@ -238,10 +238,11 @@ impl CachedRedisStorage {
         let mut interval = tokio::time::interval(flushing_period);
         tokio::spawn(async move {
             loop {
-                if p.load(Ordering::Acquire) {
-                    if storage.is_alive().await {
-                        warn!("Partition to Redis resolved!");
-                        p.store(false, Ordering::Release);
+                let partionned = !storage.is_alive().await;
```
adam-cattermole, line 241: I wonder if it'd be better renaming `partitionned` as `notAlive` or something, only because there's `partitioned`, `p`, `fn partitioned`, and now `partitionned`

I'm not tied to this though but the term feels a bit overloaded

adam-cattermole (approved):
The change makes sense to me, just one nit

limitador/src/storage/redis/redis_cached.rs
adam-cattermole: Sure I just think it's hard to read with so many instances of `partitioned`. As I've been looking at it recently it's not too difficult because I know what each variant is but if I had come to this fresh it'd be easy to miss the additional `n` being something different.

adam-cattermole: Oh I already misread it, it's `partionnned` not even `partitionned`

alexsnaps: Duh! Yes... 🤣

alexsnaps: Fixed the both types

didierofrivia:
LGTM, including the new changes that makes easier to manage the partition and simplifies within the batch update the check with just boolean instead of the Future

adam-cattermole:
Yep, the updates lgtm, makes sense to pass the bool instead of future and check if partitioned on each flushing interval.

PR #294, opened 2024-04-24
feat: CachedRedisStorage will now default to flushing a batch ASAP to…

chirino:
… reduce the length of time the cache is out of sync with Redis counters.

Setting a flushing_period is still supported to have a delay before flushing which can be useful to reduce the load on the Redis server in exchange for having staler counters in limitador.

### DIFF
--- a/limitador/src/storage/redis/redis_cached.rs
+++ b/limitador/src/storage/redis/redis_cached.rs
@@ -18,6 +18,7 @@ use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant, SystemTime};
+use tokio::sync::Notify;
 use tracing::{debug_span, error, warn, Instrument};
 
 // This is just a first version.
@@ -44,7 +45,7 @@ pub struct CachedRedisStorage {
     async_redis_storage: AsyncRedisStorage,
     redis_conn_manager: ConnectionManager,
     partitioned: Arc<AtomicBool>,
-    flush_tx: tokio::sync::watch::Sender<()>,
+    pub flush_notify: Arc<Notify>,
 }
 
 #[async_trait]
@@ -176,7 +177,7 @@ impl AsyncCounterStorage for CachedRedisStorage {
         }
 
         // ask the flusher to flush the batch
-        self.flush_tx.send(()).unwrap();
+        self.flush_notify.notify_one();
 
         Ok(Authorization::Ok)
     }
@@ -244,7 +245,7 @@ impl CachedRedisStorage {
         let batcher: Arc<Mutex<HashMap<Counter, AtomicExpiringValue>>> =
             Arc::new(Mutex::new(Default::default()));
 
-        let (flush_tx, mut flush_rx) = tokio::sync::watch::channel(());
+        let flush_notify = Arc::new(Notify::new());
 
         {
             let storage = async_redis_storage.clone();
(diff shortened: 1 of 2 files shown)


limitador/src/storage/redis/redis_cached.rs
```diff
@@ -240,15 +244,30 @@ impl CachedRedisStorage {
...
             tokio::spawn(async move {
                 loop {
+                    // Wait for a new flush request,
+                    flush_rx.changed().await.unwrap();
+
+                    if flushing_period != Duration::ZERO {
+                        // Set the flushing_period to reduce the load on Redis the downside to
+                        // setting it the cached counters will not be as accurate as when it's zero.
+                        tokio::time::sleep(flushing_period).await
+                    }
```
alexsnaps, line 265: Have you looked at tokio's [`select!`](https://docs.rs/tokio/latest/tokio/macro.select.html)? We could have either wait for the period to elapse _or_ wait to be notified... So that you'd flush _every n seconds_ or flush when notified. Cause this would have the flush task wait _after_ being notified.

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -44,6 +44,7 @@ pub struct CachedRedisStorage {
     async_redis_storage: AsyncRedisStorage,
     redis_conn_manager: ConnectionManager,
     partitioned: Arc<AtomicBool>,
+    flush_tx: tokio::sync::watch::Sender<()>,
```
alexsnaps, line 47: Wondering if a `Notify` wouldn't be more appropriate, correct me if I'm wrong, but this will enqueue every single change and have the loop on "the other side" redo all the work... while it'd race against updates in the shared DS...

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -174,6 +175,9 @@ impl AsyncCounterStorage for CachedRedisStorage {
             }
         }
 
+        // ask the flusher to flush the batch
+        self.flush_tx.send(()).unwrap();
```
alexsnaps, line 179: I like this a lot, but would make it conditional on ... being within a certain threshold of the limit? e.g. close to expiry and/or "reaching" the limit?

limitador/src/storage/redis/redis_cached.rs
chirino: Yeah it does look simpler.  Watch channels don't enqueue, they only store the last value.

alexsnaps: Ah, right, they do. And `Notify` sorts of behave similarly. So that now... with the timeout being `0` this will probably loop like crazy, other than the system being idle. The batch will be of the size of the race. So thinking we should notify "smartly"... wdyt?

limitador/src/storage/redis/redis_cached.rs
alexsnaps: See [here](https://github.com/Kuadrant/limitador/pull/294#discussion_r1579466230)... I don't know how much of an added value that notification is useful really... What do you think we're addressing by introducing this behavior? Other than the loop as fast as possible with the `0` flushing period... This "just" changes what the starting point of a flushing period is... 😕 What am I missing?

limitador/src/storage/redis/redis_cached.rs
chirino: What's the benefit of delaying the flush?  I can see the following benefits: 
1.  Reduce the number Redis API calls since you group multiple calls into 1 batch
2.  Can dedupe multiple increments of the same counter

Both of the above will likely reduce CPU load on Redi s.  But that's not the main problem global rate limiters are solving for.  The main problem is high latency calls to Redis.   Adding a delay only increases that latency further.
More latency means the local cache counters will not have the correct global value.

I do think if the load is large enough CPU load on redis will be a problem, so keeping the flushing_period option is good since that would introduce that delay that can help reduce the load on Redis.  But again, I don't think that's the main problem we are solving for.

limitador/src/storage/redis/redis_cached.rs
chirino: Updated the PR to use Notify.

limitador/src/storage/redis/redis_cached.rs
didierofrivia: The [tokio::select!](https://docs.rs/tokio/latest/tokio/macro.select.html#) would be useful if we want to _flush_ when whatever happens first, either the notify() or the flushing_period.

chirino:
yep not need any more.

Issue #298, opened 2024-04-24
Use `0` as a default on `CachedRedisStorage` cache misses


alexsnaps:
... and enqueue an eager update to the authority #296 that'd result in getting an up to date value in our cache eventually. So to avoid the RTT to the SoR

PR #299, opened 2024-04-25
Support pending writes within CachedCounterValue

alexsnaps:
Alright, this is a shot at unifying the "write-behind" and the cache we use in front of Redis.
It does so in a few ways:
 - The "pending increments" are recorded directly in the `CachedCounterValue`
 - Both the cache and the "write-behind queue" (still a Map) point to the same _instance_ of a `CachedCounterValue` (within an `Arc`) for the same `Counter` "key"
 - All entries added to the `Cache` _should be_ atomic and only then added to the `Batcher` (our write-behind "queue")
 - We _always_ look up in the `Batcher` if a `CachedCounterValue` is mapped to key, in case it got evicted from the cache
 - We _only_ add `CachedCounterValue` to the `Batcher` with an `.pending_writes() > 0`
 - We _only_ remove atomically from the `Batched` if the `.pending_writes() == 0`
 - If we ever added two different `CachedCounterValue` instances (there is a race when inserting into the cache and being evicted _before_ we made it into the `Batcher`, where we then merge the pending writes from one `CachedCounterValue` into the other one

This was achieved by doing:
 - [x] folding the pending writes within the `CachedCounterValue`
 - [x] change the `CachedRedisStorage.batcher_counter_updates` to point to the same `Arc` as within `CountersCache.cache`
 - [x] use the `CachedCounterValue` to populate the batch to redis (and update the values back)
 - [x] Introduce a proper `Batcher` type back exposing `.add()` and `.consume(u64).await`
 - [x] Have the `Batcher` awaken on either: flush period being elapsed the batch size being reached
 - [x] Flush on "important" flushes being required
 - [x] `.consume` would return the most "important" updates first always
 - [x] Lookup the queue on cache misses
 - [x] Address #298

This still misses tests 🤦 yet I still think this is better than what we had before... So, I think we should merge this and iterate, but only if my reasoning makes sense and "seems to be implemented" in this PR.

### DIFF
--- a/limitador/Cargo.toml
+++ b/limitador/Cargo.toml
@@ -22,6 +22,7 @@ lenient_conditions = []
 
 [dependencies]
 moka = { version = "0.12", features = ["sync"] }
+dashmap = "5.5.3"
 getrandom = { version = "0.2", features = ["js"] }
 serde = { version = "1", features = ["derive"] }
 postcard = { version = "1.0.4", features = ["use-std"] }
--- a/limitador/src/storage/redis/counters_cache.rs
+++ b/limitador/src/storage/redis/counters_cache.rs
@@ -9,6 +9,9 @@ use std::collections::HashMap;
 use std::sync::atomic::{AtomicI64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime};
+use tokio::select;
+use tokio::sync::Notify;
+use tokio::time::interval;
 
 pub struct CachedCounterValue {
     value: AtomicExpiringValue,
@@ -18,17 +21,36 @@ pub struct CachedCounterValue {
 
 pub struct Batcher {
     updates: Mutex<HashMap<Counter, Arc<CachedCounterValue>>>,
+    notifier: Notify,
+    interval: Duration,
 }
 
 impl Batcher {
-    fn new() -> Self {
+    fn new(period: Duration) -> Self {
         Self {
             updates: Mutex::new(Default::default()),
+            notifier: Default::default(),
+            interval: period,
         }
     }
 
     pub fn is_empty(&self) -> bool {
         self.updates.lock().unwrap().is_empty()
+    }
+
+    pub async fn consume(&self, min: usize) -> HashMap<Counter, Arc<CachedCounterValue>> {
+        let mut interval = interval(self.interval);
+        let mut ready = self.updates.lock().unwrap().len() >= min;
+        loop {
+            if ready {
+                return self.consume_all();
+            } else {
+                ready = select! {
+                    _ = self.notifier.notified() => self.updates.lock().unwrap().len() >= min,
+                    _ = interval.tick() => true,
+                }
+            }
+        }
     }
 
     pub fn consume_all(&self) -> HashMap<Counter, Arc<CachedCounterValue>> {
@@ -38,12 +60,13 @@ impl Batcher {
 
     pub fn add(&self, counter: Counter, value: Arc<CachedCounterValue>) {
         self.updates.lock().unwrap().entry(counter).or_insert(value);
+        self.notifier.notify_one();
     }
 }
 
 impl Default for Batcher {
     fn default() -> Self {
-        Self::new()
+        Self::new(Duration::from_millis(100))
     }
 }
 
@@ -164,12 +187,12 @@ impl CountersCacheBuilder {
         self
     }
 
-    pub fn build(&self) -> CountersCache {
+    pub fn build(&self, period: Duration) -> CountersCache {
         CountersCache {
             max_ttl_cached_counters: self.max_ttl_cached_counters,
             ttl_ratio_cached_counters: self.ttl_ratio_cached_counters,
             cache: Cache::new(self.max_cached_counters as u64),
-            batcher: Default::default(),
+            batcher: Batcher::new(period),
         }
     }
 }
@@ -294,7 +317,7 @@ mod tests {
             values,
         );
 
-        let cache = CountersCacheBuilder::new().build();
+        let cache = CountersCacheBuilder::new().build(Duration::default());
         cache.insert(
             counter.clone(),
             Some(10),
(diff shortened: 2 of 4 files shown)


limitador/src/storage/redis/counters_cache.rs
```diff
@@ -39,8 +42,46 @@ impl CachedCounterValue {
...
+            let writes = value - start;
+            if writes > 0 {
+                writes
+            } else {
+                value
+            }
+        };
+        match self
+            .initial_value
+            .compare_exchange(start, offset, Ordering::SeqCst, Ordering::SeqCst)
```
alexsnaps, line 71: I think this should `value`! Not offset... note to self... need to rethink this (and write a test!)

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -5,25 +5,84 @@ use crate::storage::redis::{
...
+    pub fn is_empty(&self) -> bool {
+        self.updates.lock().unwrap().is_empty()
+    }
+
+    pub async fn consume(&self, min: usize) -> HashMap<Counter, Arc<CachedCounterValue>> {
+        let mut interval = interval(self.interval);
+        let mut ready = self.updates.lock().unwrap().len() >= min;
+        loop {
+            if ready {
+                return self.consume_all();
```
alexsnaps, line 46: Here, we'd create a ordered list and return the top `min` entries... So this would also break all the pending updates in smaller batches, it'd probably also need to remove all `0` `CachedCounterValue::pending_writes`...

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -5,25 +5,84 @@ use crate::storage::redis::{
...
+use tokio::time::interval;
 
 pub struct CachedCounterValue {
     value: AtomicExpiringValue,
+    initial_value: AtomicI64,
     expiry: AtomicExpiryTime,
 }
 
+pub struct Batcher {
+    updates: Mutex<HashMap<Counter, Arc<CachedCounterValue>>>,
```
alexsnaps, line 23: This should eventually be replaced with a [concurrent HashMap](https://crates.io/crates/dashmap)

limitador/Cargo.toml
```diff
@@ -53,7 +53,7 @@ base64 = { version = "0.22", optional = true }
 
 [dev-dependencies]
 serial_test = "3.0"
-criterion = { version = "0.5.1", features = ["html_reports"] }
+criterion = { version = "0.5.1", features = ["html_reports", "async_tokio"] }
```
alexsnaps, line 56: That's a mistake 🤦

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -20,46 +22,91 @@ pub struct CachedCounterValue {
...
+                let mut result = HashMap::new();
+                for counter in &batch {
+                    let value = self.updates.get(counter).unwrap().clone();
+                    result.insert(counter.clone(), value);
+                }
+                let result = consumer(result).await;
+                for counter in &batch {
+                    self.updates
+                        .remove_if(counter, |_, v| v.no_pending_writes());
+                }
```
alexsnaps, line 88: If _after_ having flushed to redis no additional pending writes were added, we can remove these entries from the queue.

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -4,26 +4,132 @@ use crate::storage::redis::{
...
+                let mut batch = Vec::with_capacity(min);
+                let mut probably_fake = Vec::with_capacity(min);
+                for entry in &self.updates {
+                    if entry.value().value.ttl() < self.interval {
+                        batch.push(entry.key().clone());
+                        if batch.len() == min {
+                            break;
+                        }
+                    }
+                    if entry.value().expiry.duration() == Duration::from_secs(entry.key().seconds())
```
alexsnaps, line 63: This is bit of crappy hack here…

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -231,27 +211,23 @@ impl CachedRedisStorage {
...
-        let batcher: Arc<Mutex<HashMap<Counter, AtomicExpiringValue>>> =
-            Arc::new(Mutex::new(Default::default()));
 
         {
             let storage = async_redis_storage.clone();
             let counters_cache_clone = counters_cache.clone();
             let conn = redis_conn_manager.clone();
             let p = Arc::clone(&partitioned);
-            let batcher_flusher = batcher.clone();
             let mut interval = tokio::time::interval(flushing_period);
```
chirino, line 226: don't think this interval is needed anymore since the Batcher has it's own internal awaits.

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -4,26 +4,132 @@ use crate::storage::redis::{
...
+                    _ = interval.tick() => true,
+                }
+            }
+        }
+    }
+
+    pub fn add(&self, counter: Counter, value: Arc<CachedCounterValue>, priority: bool) {
+        let priority = priority || value.value.ttl() < self.interval;
+        self.updates.entry(counter).or_insert(value);
+        if priority {
```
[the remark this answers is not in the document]
alexsnaps: Yeah, I do this "hack" where I check the TTL on entry… to decide the things that are "important", but I think I should maybe just waste the additional bit…

limitador/src/storage/redis/counters_cache.rs
didierofrivia: Would it make sense to keep 2 queues within the Batcher, a _priority_ one that will gather the "hot" counters where conditions like this one applies (maybe then tag them with other priority rules) and notify or flush them asap/first ? We might not need to make decisions at "flushing"/consuming and separating into fake batchs. WDYT?

alexsnaps: I'm working on something... the thing is that "priority" is a time based thing... I'm trying to fold all that logic within `CachedCounterValue`

didierofrivia (approved):
The part that I particularly find hard/complex is the  `consume` method, that first iterates to get the hot counters that needs to flush asap, but then we iterate again to find a remainder and push it to the same batch (?), then we need to iterate again to have the final batch that we will send to the batch update fn. I only wonder if it's possible to simplify the collection we send to update... maybe in the near future we can review this. Great work! 👍🏼

PR #300, opened 2024-04-25
Only keep one map like struct to hold the coutners.

chirino:
Keeping as draft for now as this would need more work to make sure counters with pending increments destined for the next batch need to be evicted.


chirino:
closing out as #299 is much better.

PR #301, opened 2024-04-26
Refactor metrics to use alternate metrics crate

adam-cattermole:
# Changes

I've removed the original prometheus crate and added metrics, and prometheus metrics exporter. These support using the metrics API, and the exporter allows us to get a handle to be able to still use actix to export the metrics at `/metrics`.

* Metrics are configured with the help message via the `describe_*` methods on init
* They are no longer stored in the struct and used from the macro directly (due to the macro supporting labels but no method on the metrics types)
* The `prometheus_metrics` struct is still passed to `envoy_rls/server` and to the `http_api/server` to retain the config `limit_name_in_labels` for writing the metrics - not 100% sold on this solution though
* The prom metrics struct now retains a handle to the global prometheus recorder, as this can only be set once for the lifetime of the application. The tests share a single lazy static handle
* I've provided an example for adding a `datastore_partitioned` metric to the limitador library

# Verification

Checkout this branch and build server:
```
cargo build --bin limitador-server
```

Run the server:
```
./target/debug/limitador-server -vvv --grpc-reflection-service --limit-name-in-labels limitador-server/sandbox/limits.yaml redis_cached redis://127.0.0.1
```

Check metrics endpoint:
```
curl localhost:8080/metrics
# HELP limitador_up Limitador is running
# TYPE limitador_up gauge
limitador_up 1
```

# TODO
- [x] Fix tests, unable to configure the prometheus exporter

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -14,6 +14,7 @@ use crate::config::{
 use crate::envoy_rls::server::{run_envoy_rls_server, RateLimitHeaders};
 use crate::http_api::server::run_http_server;
 use crate::metrics::MetricsLayer;
+use ::metrics::histogram;
 use clap::{value_parser, Arg, ArgAction, Command};
 use const_format::formatcp;
 use limitador::counter::Counter;
@@ -256,7 +257,7 @@ fn find_first_negative_limit(limits: &[Limit]) -> Option<usize> {
 
 #[actix_rt::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let (config, prometheus_metrics) = {
+    let config = {
         let (config, version) = create_config();
         println!("{LIMITADOR_HEADER} {version}");
         let level = config.log_level.unwrap_or_else(|| {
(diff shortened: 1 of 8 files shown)


limitador-server/src/main.rs
```diff
@@ -14,6 +14,7 @@ use crate::config::{
 use crate::envoy_rls::server::{run_envoy_rls_server, RateLimitHeaders};
 use crate::http_api::server::run_http_server;
 use crate::metrics::MetricsLayer;
+use ::metrics::histogram;
```
alexsnaps, line 17: neat: `crate` prefix to stick to the pattern...

alexsnaps (approved):
🥳

limitador-server/src/main.rs
adam-cattermole: I think this is `::metrics::histogram` to match the upstream metrics crate, and not the local metrics module (used on the line above).. should probably rename the metrics module (containing the tracing metrics layer) to be clearer

alexsnaps: Oh yeah... duh! my bad. Guess the `::` threw me off, `historgram` is a macro 🤦 
Ignore me!

PR #303, opened 2024-04-30
More write behind

alexsnaps:


limitador/src/storage/redis/redis_cached.rs
```diff
@@ -339,7 +343,7 @@ async fn flush_batcher_and_update_counters<C: ConnectionLike>(
     } else {
         let updated_counters = cached_counters
             .batcher()
-            .consume(1, |counters| update_counters(&mut redis_conn, counters))
+            .consume(100, |counters| update_counters(&mut redis_conn, counters))
```
didierofrivia, line 346: I should've seen this in my previous review... mea culpa

limitador/tests/integration_tests.rs
```diff
@@ -537,7 +537,7 @@ mod test {
         }
 
         // We wait for the flushing period to pass so the counters are flushed in the cached storage
-        tokio::time::sleep(Duration::from_millis(3)).await;
+        tokio::time::sleep(Duration::from_millis(4)).await;
```
didierofrivia, line 540: I didn't fix this in my previous PR since I couldn't find an easy way to do so, we could tackle this later to avoid sleeping in tests

didierofrivia (approved):
Easier to read and to follow what's about.

PR #304, opened 2024-04-30
fixes #295: Use a semaphore to protect the Batcher from unbounded memory growth.

### LINKED ISSUE #295: Do not OOM due to `CountersCache.Batcher` getting too big
We need to either explicitly bound that data structure or bound it thru heuristics. 
Today, it'll grow forever, most importantly when partitioned... It could certainly be compacted (e.g. expired `AtomicExpiringValue`) it is mostly the key space that's the issue, as it is derived from the user's defined limits and could be really big (e.g. many conditions, large variables, ...)

chirino:


limitador/src/storage/redis/counters_cache.rs
```diff
@@ -141,10 +143,11 @@ impl Batcher {
             notifier: Default::default(),
             interval: period,
             priority_flush: AtomicBool::new(false),
+            limiter: Semaphore::new(1000),
```
chirino, line 146: we may want to make this configurable later.  Is there a better default we can use for now?

alexsnaps: Or we'd need to monitor memory usage and do it all conditional on that... Not sure I want to start going down that rabbit hole tho 🙈

alexsnaps: Maybe we could, for now, use the same [`max_cached_counters` value](https://github.com/Kuadrant/limitador/blob/65d34f15a35240c9ffdf2a0d0609b68b74dde45e/limitador/src/storage/redis/counters_cache.rs#L379-L380). Gives the user some control, while avoiding leaking too many "implementation details" just quiet yet, wdyt?

chirino: Sounds good.  Updated.

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -189,8 +193,13 @@ impl Batcher {
...
                 let result = consumer(result).await;
                 batch.iter().for_each(|counter| {
-                    self.updates
-                        .remove_if(counter, |_, v| v.no_pending_writes());
+                    if self
+                        .updates
+                        .remove_if(counter, |_, v| v.no_pending_writes())
+                        .is_some()
+                    {
+                        self.limiter.add_permits(1);
```
alexsnaps, line 201: I think the example in the doc is somewhat oversimplified...
Reading the docs, `.add_permits` doesn't account for the initial permits provided (i.e. consider it a max), see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=42c8b23185148572e6111882673eb2c4) 
So we need to do all the math ourselves, as this wouldn't account for re-adding permits back when counters expire for instance. I'd expect this would run out of permit after a while.

alexsnaps (changes requested):
We can't use `Semaphore` that easily I fear. So either we move to a different notification model, or we do the proper math (tho that'd be racy) on adding permits back based off the size of the pending writes and the permits remaining.

limitador/src/storage/redis/counters_cache.rs
alexsnaps: Actually... thinking about it, it might be good enough. As we don't compact the queue anywhere 🤔 
So that, if the entry is passed to the `consumer`, even if there are no pending writes for this entry e.g. because it expired, then this would also remove it from `self.updates` and increment the tickets back...

PR #305, opened 2024-04-30
PRs comming from forks do not push images

eguzki:
### What

GH action runs on pushes coming from forked PR's do not have permissions to access docker image repo in quay.io.

This PR adds some conditionals to skip those steps that would fail otherwise.

But building the image and running smoke tests on the image is still a value for forked PR's, so allowing them as a separated job. Not doing multi-arch build with layer caching as the regular image building, though.

### Tested

- [ ] Pushes from kuadrant users
- [ ] Pushes from external users (forked PR's)
- [ ] Tags

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -133,7 +133,7 @@ jobs:
     name: Build on forks
     # for forks
     if: |
-      github.event.repository.full_name != github.repository)
+      github.event.repository.full_name != github.repository
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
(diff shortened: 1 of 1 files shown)


alexsnaps (approved):
Yeah this should work... I was a bit fast at merging #306 - it doesn't fix it. Probably need to revert it... my bad

eguzki:
Currenlty, the even trigger configuration prevents image building from PRs from forked repos. 

```yaml
on:
  workflow_dispatch:
  push:
    branches: ['*']
    tags: ['*']
```
the `push` event is never triggered on PR's from forked repos because there is no new commits/tags in the kuadrant's repo, but instead in another repo.

PR #309, opened 2024-05-01
feat: add support for getting DraftVersion03 response headers on the /check_and_report operation.

chirino:


limitador-server/docs/http_server_spec.json
```diff
@@ -16,6 +16,10 @@
           "additionalProperties": {
             "type": "string"
           }
+        },
+        "response_headers": {
```
alexsnaps, line 20: Is this supposed to be camelCase? There seems to be a mix here already… `CheckAndReportInfo` vs `additionalProperties` (lowercase)… Don't know it there is a recommendation for either/or but camel & snake seems weird, no?

alexsnaps (approved):

limitador-server/docs/http_server_spec.json
chirino: additionalProperties is not the name of a field... it's used to define the type of the key for dictionary types.  see: https://swagger.io/docs/specification/data-models/dictionaries/ 

The norm is CamelCase type names but snake case field names. Here is another field that is using sake case: https://github.com/Kuadrant/limitador/blob/b2695aec5e7ab128420a4a694ab968d29e840f3d/limitador-server/docs/http_server_spec.json#L80

chirino:
rebased

chirino:
rebased

PR #310, opened 2024-05-01
Added tests... and fixes... to CachedCounterValue

alexsnaps:
WIP, but... 🤦 since there are bugs, lets make this a first batch of more tests to come


adam-cattermole (approved):
Changes lgtm

PR #314, opened 2024-05-02
Easier expiry

alexsnaps:
this change is about _not_ reset a `CachedCounterValue` when it's expired, but rather keep its value and consider it requiring a fast flush (which it already did before, by considering it as if it was absent from the cache for the `check_and_update` body, but then `increase_by` would still increment the expired value).
This now uses the value from the `cache` the entire call stack...

### DIFF
--- a/limitador/src/storage/redis/counters_cache.rs
+++ b/limitador/src/storage/redis/counters_cache.rs
@@ -128,7 +128,9 @@ impl CachedCounterValue {
     }
 
     pub fn requires_fast_flush(&self, within: &Duration) -> bool {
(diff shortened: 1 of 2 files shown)


adam-cattermole (commented):
I think this makes sense to me. We set the priority if the cached value has expired instead of treating it as uncached. In my minor bit of testing though once the cached value expired I saw that this expired_at was always returning true which makes me wonder why the cached counter expiry.. perhaps my running conditions?

adam-cattermole (approved):
There was a mistake in my testing setup, checked on top of changes in 316 and working as I'd expect

didierofrivia (approved):
🍪

PR #315, opened 2024-05-03
Batcher tests

alexsnaps:


adam-cattermole (approved):
Tests all make sense, lgtm

PR #316, opened 2024-05-07
Cached delta updates

alexsnaps:
Waiting on #315 to be reviewed and merged

### DIFF
--- a/limitador/src/storage/redis/redis_cached.rs
+++ b/limitador/src/storage/redis/redis_cached.rs
@@ -292,43 +292,48 @@ impl CachedRedisStorageBuilder {
 async fn update_counters<C: ConnectionLike>(
     redis_conn: &mut C,
     counters_and_deltas: HashMap<Counter, Arc<CachedCounterValue>>,
-) -> Result<Vec<(Counter, i64, i64)>, StorageErr> {
+) -> Result<Vec<(Counter, i64, i64, i64)>, StorageErr> {
     let redis_script = redis::Script::new(BATCH_UPDATE_COUNTERS);
     let mut script_invocation = redis_script.prepare_invoke();
 
-    let mut res: Vec<(Counter, i64, i64)> = Vec::with_capacity(counters_and_deltas.len());
-    if counters_and_deltas.is_empty() {
-        return Ok(res);
-    }
+    let res = if counters_and_deltas.is_empty() {
+        Default::default()
+    } else {
+        let mut res: Vec<(Counter, i64, i64, i64)> = Vec::with_capacity(counters_and_deltas.len());
 
-    for (counter, value) in counters_and_deltas {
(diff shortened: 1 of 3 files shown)


limitador/src/storage/redis/redis_cached.rs
```diff
@@ -292,43 +292,48 @@ impl CachedRedisStorageBuilder {
 async fn update_counters<C: ConnectionLike>(
     redis_conn: &mut C,
     counters_and_deltas: HashMap<Counter, Arc<CachedCounterValue>>,
-) -> Result<Vec<(Counter, i64, i64)>, StorageErr> {
+) -> Result<Vec<(Counter, i64, i64, i64)>, StorageErr> {
```
didierofrivia, line 295: the new value corresponds to the delta in the last position? maybe we could add an doc annotation to this fn

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -292,43 +292,48 @@ impl CachedRedisStorageBuilder {
...
+
+        // We need to update the values and ttls returned by redis
+        let counters_range = 0..res.len();
+        let script_res_range = (0..script_res.len()).step_by(2);
+
+        for (i, j) in counters_range.zip(script_res_range) {
+            let (_, val, delta, ttl) = &mut res[i];
+            *val = script_res[j];
+            *delta = script_res[j] - *delta;
+            *ttl = script_res[j + 1];
```
didierofrivia, line 333: Ok, so last one is ttl

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -438,21 +447,22 @@ mod tests {
...
 
-        let (c, v, t) = result.unwrap()[0].clone();
+        let (c, new_value, remote_increments, new_ttl) = result.remove(0);
+        assert_eq!(key_for_counter(&counter), key_for_counter(&c));
+        assert_eq!(NEW_VALUE_FROM_REDIS, new_value);
         assert_eq!(
-            "req.method == \"GET\"",
-            c.limit().conditions().iter().collect::<Vec<_>>()[0]
+            NEW_VALUE_FROM_REDIS - INITIAL_VALUE_FROM_REDIS - LOCAL_INCREMENTS,
+            remote_increments
```
didierofrivia, line 463: explicit, so we can make assertions with the redis deltas 👍🏼

didierofrivia (approved):
🍸

limitador/src/storage/redis/redis_cached.rs
alexsnaps: I'll burn the whole ttl stuff, use absolute timestamps instead and then refactor this... 
I hate tuples... of 4 elements even more so.

PR #317, opened 2024-05-08
Use absolute timestamps when dealing with expiry from Redis

alexsnaps:


limitador/src/storage/redis/scripts.rs
```diff
@@ -38,10 +38,8 @@ pub const BATCH_UPDATE_COUNTERS: &str = "
         if c == tonumber(delta) then
             redis.call('expire', counter_key, ttl)
             redis.call('sadd', limit_key, counter_key)
-            table.insert(res, ttl*1000)
-        else
-            table.insert(res, redis.call('pttl', counter_key))
         end
+        table.insert(res, redis.call('pexpiretime', counter_key))
```
alexsnaps, line 42: This isn't ideal, as `pexpiretime` was introduced in Redis 7... that being said, [elasticache](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/SupportedCommands.html) supports it

adam-cattermole: Not that it matters right now but it's also supported by [memorystore](https://cloud.google.com/memorystore/docs/cluster/supported-commands) and I think Azure cache

limitador/src/storage/redis/counters_cache.rs
```diff
@@ -260,30 +245,24 @@ impl CountersCache {
...
-        let cache_ttl = self.ttl_from_redis_ttl(
-            redis_ttl_ms,
-            counter.seconds(),
-            redis_val,
-            counter.max_value(),
-        );
-        if let Some(ttl) = cache_ttl.checked_sub(ttl_margin) {
-            if ttl > Duration::ZERO {
+        if redis_expiry > 0 {
+            let expiry_ts = SystemTime::UNIX_EPOCH + Duration::from_micros(redis_expiry as u64);
```
adam-cattermole, line 251: According to the docs for [PEXPIRETIME](https://redis.io/docs/latest/commands/pexpiretime/) it's returned as millis, not micros so I see the `if expiry_ts > SytemTime::now()` returning false when it's supposed to be true

```suggestion
            let expiry_ts = SystemTime::UNIX_EPOCH + Duration::from_millis(redis_expiry as u64);
```

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -326,10 +315,10 @@ async fn update_counters<C: ConnectionLike>(
         let script_res_range = (0..script_res.len()).step_by(2);
 
         for (i, j) in counters_range.zip(script_res_range) {
-            let (_, val, delta, ttl) = &mut res[i];
+            let (_, val, delta, expireat) = &mut res[i];
```
didierofrivia, line 306: maybe to keep consistency should we rename `expireat` to `expires_at` (?)

limitador/src/storage/redis/counters_cache.rs
alexsnaps: Scares me this didn't create any test failures! 🤦

alexsnaps: Ah! Because I had written the tests to inject time in micros too… fixed that part as well.

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -362,18 +339,8 @@ async fn flush_batcher_and_update_counters<C: ConnectionLike>(
...
-            cached_counters.apply_remote_delta(
-                counter,
-                new_value,
-                remote_deltas,
-                ttl,
-                Duration::from_millis(
-                    (Instant::now() - time_start_update_counters).as_millis() as u64
-                ),
-            );
+            cached_counters.apply_remote_delta(counter, new_value, remote_deltas, ttl);
```
adam-cattermole, line 343: nit: maybe we should also change this `ttl` to `expires_at`

adam-cattermole (approved):
Updates look good, nice one

PR #318, opened 2024-05-08
Batch size

alexsnaps:
Based off #317

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -24,8 +24,8 @@ use limitador::storage::disk::DiskStorage;
 #[cfg(feature = "infinispan")]
 use limitador::storage::infinispan::{Consistency, InfinispanStorageBuilder};
 use limitador::storage::redis::{
-    AsyncRedisStorage, CachedRedisStorage, CachedRedisStorageBuilder, DEFAULT_FLUSHING_PERIOD_SEC,
-    DEFAULT_MAX_CACHED_COUNTERS, DEFAULT_RESPONSE_TIMEOUT_MS,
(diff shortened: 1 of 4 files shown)


limitador-server/src/main.rs
```diff
@@ -588,6 +589,18 @@ fn create_config() -> (Configuration, &'static str) {
...
+                .arg(
+                    Arg::new("batch")
+                        .long("batch-size")
+                        .action(ArgAction::Set)
+                        .value_parser(clap::value_parser!(u64))
+                        .default_value(
+                            config::env::REDIS_LOCAL_CACHE_BATCH_SIZE
+                                .unwrap_or(leak(DEFAULT_BATCH_SIZE)),
+                        )
+                        .display_order(3)
```
adam-cattermole, line 601: I actually don't know if it matters or why it was starting from 4 before, but shouldn't these sub-options start at 1 through 4.. although I guess as long as they can be ordered it doesn't matter

adam-cattermole (commented):
Should we update the doc as well.. remove the old ttl vars and add this batch size one https://github.com/Kuadrant/limitador/blob/main/doc/server/configuration.md?plain=1#L216-L232
https://github.com/Kuadrant/limitador/blob/main/doc/server/configuration.md?plain=1#L385-L400

limitador-server/src/main.rs
alexsnaps: It's "just" a way to sort the options, by avoiding changing the order... e.g. to avoid crappy diffs. Which is why I didn't change these here neither and keep the diff to the "useful" stuff

adam-cattermole (approved):
Happy to approve, given the docs are already out of date we can update in a follow up if preferred

didierofrivia (approved):
🍰

PR #319, opened 2024-05-09
Revert on transient redis error

### LINKED ISSUE #297: `async fn flush_batcher_and_update_counters` drops pending writes on partition
If during a flush an error occurs that has Limitador flip the `partitioned` flag, all these pending writes are dropped on the floor. 
Additionally, a timeout will probably occur on large enough batches... we'd be better off using a different timeout for these updates

alexsnaps:
Fixes #297 
Based off #318 …

### DIFF
--- a/limitador/src/storage/redis/counters_cache.rs
+++ b/limitador/src/storage/redis/counters_cache.rs
@@ -104,6 +104,22 @@ impl CachedCounterValue {
         value - start == 0
     }
 
+    fn revert_writes(&self, writes: u64) -> Result<(), ()> {
+        let newer = self.initial_value.load(Ordering::SeqCst);
+        if newer > writes {
+            return match self.initial_value.compare_exchange(
+                newer,
+                newer - writes,
+                Ordering::SeqCst,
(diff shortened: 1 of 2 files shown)


limitador/src/storage/redis/counters_cache.rs
```diff
@@ -160,10 +176,10 @@ impl Batcher {
         self.notifier.notify_one();
     }
 
-    pub async fn consume<F, Fut, O>(&self, max: usize, consumer: F) -> O
+    pub async fn consume<F, Fut, T, E>(&self, max: usize, consumer: F) -> Result<T, E>
     where
         F: FnOnce(HashMap<Counter, Arc<CachedCounterValue>>) -> Fut,
-        Fut: Future<Output = O>,
+        Fut: Future<Output = Result<T, E>>,
```
alexsnaps, line 182: This signature is getting somewhat … ugly.  But `async` closures aren't a stable thing yet, so I can't have anything call `await` in the closure passed to `consume`… that's how I went about it instead…

PR #324, opened 2024-05-14
Crdt

alexsnaps:
Creating this here, if someone wants to play with this... 
I think this exposes all seams to share using the `CrCounterValue`s over gRPC streams

### DIFF
--- a/limitador/src/storage/distributed/mod.rs
+++ b/limitador/src/storage/distributed/mod.rs
@@ -1,22 +1,30 @@
-mod cr_counter_value;
+use std::collections::hash_map::Entry;
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::net::ToSocketAddrs;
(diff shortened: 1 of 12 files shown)


limitador/src/storage/distributed/mod.rs
```diff
@@ -0,0 +1,429 @@
...
+                    if addr != local {
+                        match postcard::from_bytes::<CounterValueMessage>(&buf[..len]) {
+                            Ok(message) => {
+                                let CounterValueMessage {
+                                    counter_key,
+                                    expiry,
+                                    values,
+                                } = message;
+                                let counter = <CounterKey as Into<Counter>>::into(counter_key);
+                                let counters = limits_for_namespace.read().unwrap();
```
alexsnaps, line 300: This should check for qualified counters as well...!

chirino (approved):
Lgtm

PR #339, opened 2024-05-21
All the fixes

alexsnaps:
If you test this in `INFO` level log as such: 

```shell
$ ./target/debug/limitador-server limitador-server/examples/limits.yaml -vv --grpc-reflection-service redis_cached  redis://localhost:6379
```

And then start the driver, you'd get a first log for the priority flush: 

```
INFO flush_batcher_and_update_counters: limitador::storage::redis::counters_cache: Priority flush!
```

Then further for the per second flushes: 

```
2024-05-21T15:43:58.456434Z  INFO flush_batcher_and_update_counters: limitador::storage::redis::redis_cached: Flushing 1 counter updates
2024-05-21T15:43:59.458124Z  INFO flush_batcher_and_update_counters: limitador::storage::redis::redis_cached: Flushing 1 counter updates
```

Then shut redis down: 

```
2024-05-21T15:44:03.472454Z ERROR flush_batcher_and_update_counters: limitador::storage::redis::redis_cached: Partition to Redis detected!
2024-05-21T15:44:03.472467Z  WARN flush_batcher_and_update_counters: limitador::storage::redis::redis_cached: Error flushing error while accessing the limits storage: broken pipe
```

No more logs for the duration of the partition, restart redis, and...

```
2024-05-21T15:44:08.480444Z  WARN flush_batcher_and_update_counters: limitador::storage::redis::redis_cached: Partition to Redis resolved!
2024-05-21T15:44:09.481024Z  INFO flush_batcher_and_update_counters: limitador::storage::redis::redis_cached: Flushing 1 counter updates
```

### DIFF
--- a/limitador/src/storage/redis/redis_async.rs
+++ b/limitador/src/storage/redis/redis_async.rs
@@ -9,12 +9,11 @@ use crate::storage::redis::is_limited;
 use crate::storage::redis::scripts::{SCRIPT_UPDATE_COUNTER, VALUES_AND_TTLS};
 use crate::storage::{AsyncCounterStorage, Authorization, StorageErr};
 use async_trait::async_trait;
-use metrics::histogram;
 use redis::{AsyncCommands, RedisError};
 use std::collections::HashSet;
 use std::str::FromStr;
-use std::time::{Duration, Instant};
-use tracing::{debug_span, warn, Instrument};
+use std::time::Duration;
+use tracing::{debug_span, Instrument};
 
 // Note: this implementation does not guarantee exact limits. Ensuring that we
 // never go over the limits would hurt performance. This implementation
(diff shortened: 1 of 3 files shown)


limitador/src/storage/redis/redis_async.rs
```diff
@@ -215,6 +215,14 @@ impl AsyncRedisStorage {
             .clone()
             .incr::<&str, i32, u64>("LIMITADOR_LIVE_CHECK", 1)
             .await
+            .map_err(|err| {
```
alexsnaps, line 218: This `is_alive(&self)` isn't used anymore afaict... but there is the `.then` wrt liveness_latency... So... what shall we do?

limitador/src/storage/redis/redis_cached.rs
```diff
@@ -339,48 +338,51 @@ async fn update_counters<C: ConnectionLike>(
...
+        .batcher()
+        .consume(batch_size, |counters| {
+            if !counters.is_empty() && !partitioned.load(Acquire) {
+                info!("Flushing {} counter updates", counters.len());
+            }
+            update_counters(&mut redis_conn, counters)
+        })
+        .await
+        .map(|res| {
+            // info!("Success {} counters", res.len());
```
didierofrivia, line 355: 🔥

limitador/src/storage/redis/redis_async.rs
didierofrivia: I haven't found a ` #[allow(dead_code)] ` that might affect this (not complaining), also, method `is_alive` is not found in any other part of the project than here (method definition)

alexsnaps: @adam-cattermole what you recon we do? 🔥 the metric? Call it from a dedicated `Future`?

didierofrivia (approved):
Everything looks good to me, with the exception of cargo not complaining for the unused method `is_alive`. Maybe @adam-cattermole would have a saying on this?

limitador/src/storage/redis/redis_async.rs
adam-cattermole: I reckon we just burn the metric/method. We can add this back if needed but for now we could just use the `datastore_latency` from the batcher calls to redis (I know even this metric isn't perfect due to size of batch etc). If we decide we want this down the line it shouldn't be too hard to add it back

adam-cattermole (approved):
lgtm

Issue #338, opened 2024-05-20
CI: cache rust bits for faster CI build.


chirino:
Consider using https://github.com/Swatinem/rust-cache github action to speed up the rust builds.

PR #341, opened 2024-05-22
GitHub CI improvements.

chirino:
* Only run the image build on main and tags (not needed on feature branches).
* Cache rust dependencies to speed builds up.


chirino:
This PR fixes #338

PR #342, opened 2024-05-23
Fix `--no-default-features` and bench

alexsnaps:
`redis-test` in our dev-dependencies depends on `redis` but doesn't enable the async runtime... so this needs to be done explicitly by our crate.

In turn: `--no-default-features` compiles _and_ lets us run `bench` on just the in mem store, or that and other `*_storage`s as one decides


chirino (approved):
lgtm

PR #343, opened 2024-05-23
[distributed storage] support re-syncing counters after a peer is partitioned.

chirino:
This likely needs more thought put into it.  A peer only resyncs when it goes from having zero active connections to 1 active connection.

### DIFF
--- a/limitador/src/storage/distributed/grpc/mod.rs
+++ b/limitador/src/storage/distributed/grpc/mod.rs
@@ -5,10 +5,10 @@ use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{error::Error, io::ErrorKind, pin::Pin};
 
+use tokio::sync::mpsc::error::TrySendError;
 use tokio::sync::mpsc::Sender;
 use tokio::sync::{broadcast, mpsc, RwLock};
 use tokio::time::sleep;
-
 use tokio_stream::{wrappers::ReceiverStream, Stream, StreamExt};
 use tonic::{Code, Request, Response, Status, Streaming};
 use tracing::debug;
@@ -17,7 +17,7 @@ use crate::storage::distributed::grpc::v1::packet::Message;
 use crate::storage::distributed::grpc::v1::replication_client::ReplicationClient;
 use crate::storage::distributed::grpc::v1::replication_server::{Replication, ReplicationServer};
 use crate::storage::distributed::grpc::v1::{
-    CounterUpdate, Hello, MembershipUpdate, Packet, Peer, Pong,
+    CounterUpdate, Empty, Hello, MembershipUpdate, Packet, Peer, Pong,
 };
 
 // clippy will barf on protobuff generated code for enum variants in
@@ -84,8 +84,8 @@ struct Session {
 
 impl Session {
     async fn close(&mut self) {
-        let mut state = self.replication_state.write().await;
-        if let Some(peer) = state.peer_trackers.get_mut(&self.peer_id) {
+        let mut replication_state = self.replication_state.write().await;
+        if let Some(peer) = replication_state.peer_trackers.get_mut(&self.peer_id) {
             peer.session = None;
         }
     }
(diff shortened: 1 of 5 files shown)


chirino:
Trying to avoid doing too many re-sync on every connect to a peer, but I think we could end up in scenarios where you have peers A, B, C, D and
peer A and B start up and connect and sync,
then C and D start up and connect and sync, 
then C,D and A,B connect and don't sync anymore as they thought they were already synced with the cluster.

chirino:
Switched to a more simplified strategy of syncing with every peer session.  It should work better in the face of complex network partitions.

limitador/src/storage/distributed/mod.rs
```diff
@@ -279,6 +296,49 @@ impl CrInMemoryStorage {
     }
 }
 
+async fn process_re_sync(
```
alexsnaps, line 299: Is the contract that this gets called _prior_ receiving any remote updates?
Not that it would matter in terms of correctness, but... see [below](https://github.com/Kuadrant/limitador/pull/343/files#diff-c3ae63bee7abe7f0e5a1c02891d517102a556f6855d8df57c7c654f29a3c05efR316)

limitador/src/storage/distributed/mod.rs
```diff
@@ -241,6 +247,17 @@ impl CrInMemoryStorage {
             });
         }
 
+        // process the re-sync requests...
+        {
+            let limits = limits.clone();
+            tokio::spawn(async move {
+                let limits = limits.clone();
```
alexsnaps, line 254: Confused why we `clone()` the `limits` twice here, am I misreading this?

limitador/src/storage/distributed/mod.rs
alexsnaps: Ok, no... I misread this, this _sends_ the data to a newly connected peer! right?

limitador/src/storage/distributed/mod.rs
```diff
@@ -279,6 +296,49 @@ impl CrInMemoryStorage {
...
+        limits.keys().cloned().collect()
+    };
+
+    for key in keys {
+        let update = {
+            let limits = limits.read().unwrap();
+            limits.get(&key).and_then(|store_value| {
+                let (expiry, ourself, value) = store_value.clone().into_ourselves_inner();
+                if value == 0 {
+                    None // no point in sending a counter that is empty
```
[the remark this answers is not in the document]
alexsnaps: Because, that's a send, you decided that each node would only send the counter that they have a value for "themselves" and let the others do the same for _their own values_? Am I now understanding this right?

limitador/src/storage/distributed/mod.rs
chirino: yout right.. not needed.

limitador/src/storage/distributed/mod.rs
chirino: correct..  also note that we only re-sync 1 peer at a time since the requests to resync the peer are delivered on a channel service by 1 task... not sure if that's good or not.  I was thinking that you might want to avoid impacting this node too much by other peers connecting to it.  Maybe what would be ideal is resyncing up to X concurrent peers.

limitador/src/storage/distributed/mod.rs
```diff
@@ -279,6 +296,49 @@ impl CrInMemoryStorage {
...
+    let keys: Vec<_> = {
+        let limits = limits.read().unwrap();
+        limits.keys().cloned().collect()
+    };
+
+    for key in keys {
+        let update = {
+            let limits = limits.read().unwrap();
+            limits.get(&key).and_then(|store_value| {
+                let (expiry, ourself, value) = store_value.clone().into_ourselves_inner();
```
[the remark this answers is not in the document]
chirino: was trying to avoid having to clone the others, but the caller was doing anyways.  Going to re-write that bit.

limitador/src/storage/distributed/mod.rs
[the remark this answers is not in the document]
chirino: Correct.  But I think we should also check the expired condition, so I'll update that.

alexsnaps (approved):
Thanks. I'll try to wrap my head around all of this "better" next week. 
But there is no world in which all of this PR isn't making this better already!

limitador/src/storage/distributed/grpc/mod.rs
```diff
@@ -123,11 +162,11 @@ impl Session {
                             self.process_packet(packet).await?;
                         },
                         Some(Err(err)) => {
-                            if is_disconnect(&err) {
+                            return if is_disconnect(&err) {
```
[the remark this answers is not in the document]
alexsnaps: ... and if we "clean" disconnect, we'll also reconnect cleanly later I guess... makes sense.

limitador/src/storage/distributed/mod.rs
alexsnaps: 👌 let's iterate again, we'll test these things out in "production" scenarios and improvise from actual data.

PR #344, opened 2024-05-23
in mem storage: got rid of the namespace indirection

alexsnaps:
image

Looks better _in this screenshot_, but I will flamegraph and test this a little further on linux

### DIFF
--- a/limitador/src/storage/in_memory.rs
+++ b/limitador/src/storage/in_memory.rs
@@ -3,14 +3,14 @@ use crate::limit::{Limit, Namespace};
 use crate::storage::atomic_expiring_value::AtomicExpiringValue;
 use crate::storage::{Authorization, CounterStorage, StorageErr};
 use moka::sync::Cache;
-use std::collections::hash_map::Entry;
-use std::collections::{HashMap, HashSet};
+use std::collections::btree_map::Entry;
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::ops::Deref;
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, SystemTime};
 
 pub struct InMemoryStorage {
-    simple_limits: RwLock<HashMap<Limit, AtomicExpiringValue>>,
+    simple_limits: RwLock<BTreeMap<Limit, AtomicExpiringValue>>,
     qualified_counters: Cache<Counter, Arc<AtomicExpiringValue>>,
 }
 
@@ -197,7 +197,7 @@ impl CounterStorage for InMemoryStorage {
 impl InMemoryStorage {
     pub fn new(cache_size: u64) -> Self {
         Self {
-            simple_limits: RwLock::new(HashMap::new()),
+            simple_limits: RwLock::new(BTreeMap::new()),
             qualified_counters: Cache::new(cache_size),
         }
     }
--- a/limitador/src/storage/in_memory.rs
+++ b/limitador/src/storage/in_memory.rs
@@ -9,29 +9,26 @@ use std::ops::Deref;
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, SystemTime};
 
-type NamespacedLimitCounters<T> = HashMap<Namespace, HashMap<Limit, T>>;
-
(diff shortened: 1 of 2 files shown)


chirino (approved):
lgtm

alexsnaps:
image
Some further testing, with no `variables`, i.e actually hitting the change! Not that much difference... With the newly added bench scenario.

But if I grow the namespaces to a `1000`... 
image
Then it's a regression!

Now, that's all still on MacOS, I'll test these more rigorously on Linux next...

alexsnaps:
Also... if a "tree" (`namespace` 1 -> * `Limits` -> 1 `Counter`) seem to be faster... how would storing these in a `BTreeMap` work our... 🤔 
image
Well much... better. Anyways, Imma stop here and run all these on Linux and flamegraph some of these, but some of this would absolutely apply to our distributed storage /cc @chirino

limitador/src/storage/in_memory.rs
```diff
@@ -9,49 +9,42 @@ use std::ops::Deref;
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, SystemTime};
 
-type NamespacedLimitCounters<T> = HashMap<Namespace, HashMap<Limit, T>>;
-
 pub struct InMemoryStorage {
-    limits_for_namespace: RwLock<NamespacedLimitCounters<AtomicExpiringValue>>,
+    simple_limits: RwLock<HashMap<Limit, AtomicExpiringValue>>,
```
didierofrivia: ❤️

limitador/src/storage/in_memory.rs
```diff
@@ -3,14 +3,14 @@ use crate::limit::{Limit, Namespace};
...
-use std::collections::{HashMap, HashSet};
+use std::collections::btree_map::Entry;
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::ops::Deref;
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, SystemTime};
 
 pub struct InMemoryStorage {
-    simple_limits: RwLock<HashMap<Limit, AtomicExpiringValue>>,
+    simple_limits: RwLock<BTreeMap<Limit, AtomicExpiringValue>>,
```
didierofrivia, line 13: 👍🏼

alexsnaps:
Did some more and, by now, relatively extensive testing... having the `Limit`s comparable and store `AtomicExpiringValue` in a `BTreeMap` is giving the best results in most of the cases. When you have `Limit`s with _many_ (in the 100s) of  `Condition`s, performance degrades. Still need to look into that further... But here is the callstack for `Hash` based vs `Cmp` based _when searching for the `Counter`s_:

## Hash based: `HashMap<Namespace, HashMap<Limit, AtomicExpiringValue>>`
Screenshot from 2024-05-27 16-21-23


## Cmp based: `BTreeMap<Limit, AtomicExpiringValue>`

Screenshot from 2024-05-27 16-23-13


Giving us this kind of results:

```
Memory/check_rate_limited_and_update/10 namespaces with 50 limits each with 10 conditions and 0 vari...
                        time:   [253.02 µs 254.21 µs 255.86 µs]
                        change: [-9.1814% -8.4315% -7.7729%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Screenshot from 2024-05-27 16-34-04


## Zooming out...

Right now the worst offender tho, is not where we are looking for, in which ever storage version here btw:
Screenshot from 2024-05-27 16-27-57

`RateLimiter::counters_that_applies` spends all its time `clone()` the `Limit`s (and composing parts) to then re-clone again and map into `Counter`s... I'll have a look at that next.

alexsnaps:
>🛑 This cannot be merged! ⚠️
>
>As of today we use the JSON serialized form of Limits to look counters up in Redis... It is highly unlikely that switching from >HashMaps for our Limit field to BTreeMap would preserve the order we obtained in our own sorting functions.

I'll break the format anyways with more PRs coming… which is why we are releasing Limitador `v2`, with that in mind, maybe we are ok merging this?

didierofrivia (approved):
🪜

PR #345, opened 2024-05-24
[distributed store] Batch up updates per session.

chirino:


chirino:
rebased off main

alexsnaps (approved):
I think this could be "simplified" from a "design" perspective and as such avoid some `.clone()` et al... Have a slightly more "fluid" data flow. But I could also be _wrong_. Let's keep on moving forward and revisit when we have a better overview of the whole solution _and have our back covered with more tests_ :)

PR #346, opened 2024-05-27
Added test scenario for no variable

alexsnaps:
i.e. no qualified counters for the limits


alexsnaps:
Seems safe enough... Merging

PR #348, opened 2024-05-28
Refactor levels for metrics, logging, and tracing

adam-cattermole:
## Changes

- Apply level filter explicitly to the `fmt_layer` (logging layer)
- Now sets tracing level to `INFO` minimum, however still configurable by verbosity to `DEBUG`/`TRACE`

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -43,6 +43,7 @@ use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 use std::{env, process};
+use tracing_subscriber::Layer;
 
 #[cfg(feature = "distributed_storage")]
 use clap::parser::ValuesRef;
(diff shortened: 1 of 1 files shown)


limitador-server/src/main.rs
```diff
@@ -251,16 +252,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
             tracing_subscriber::registry()
-                .with(level)
                 .with(metrics_layer)
                 .with(fmt_layer)
+                .with(tracing_level)
                 .with(telemetry_layer)
                 .init();
         } else {
             tracing_subscriber::registry()
-                .with(level)
```
adam-cattermole, line 263: Worth mentioning that removing this global level filter means all spans hit the metrics layer regardless of the configured verbosity now

alexsnaps (approved):
I guess you didn't perf profile this? I realize this would require a new bench...

adam-cattermole:
I've done some initial investigation into the performance effects of this and we came to the conclusion that this is acceptable so going to merge.

Going to continue to performance profile and will provide a further PR if I can determine any potential performance improvements

PR #350, opened 2024-06-03
Simplify image builds by using QEMU to do multi-arch builds

chirino:
but since they are slow, only do the multi-arch builds for releases.

Also support passing cargo build args to the Docker image builder, allowing us to enable all features.


.github/workflows/build-image.yaml
```diff
@@ -6,8 +6,6 @@ on:
   push:
     branches:
       - main
-    tags:
-      - "*"
```
alexsnaps, line 10: I think we need this, as that's how we release, no? @didierofrivia 
tags could be only `v[0-9]+\.[0-9]+\.[0-9]+(-.*)?` or something tho

.github/workflows/build-multi-arch-image.yaml
```diff
@@ -0,0 +1,28 @@
+---
+name: Build Image
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - "*"
```
chirino, line 8: Added another workflow for releases, in this case we build both archs.

eguzki:
If I understand this correctly, image is not being built on PR's. Is this something we want? Building an image is kind of a test as well that either passes or fails.

eguzki:
> If I understand this correctly, image is not being built on PR's. Is this something we want? Building an image is kind of a test as well that either passes or fails.

I realize that the image building was removed from PR's a while ago.

didierofrivia:
I'd say that going back to quemu for the default building is not ideal as the default build job, I'd rather add it as a different `dispatch` workflow and/or only for releases

chirino:
ok.  closing PR, maybe revisit later.

PR #351, opened 2024-06-04
Add some load testing tools to the sandbox

chirino:
Documents how to use ghz to benchmark the gRPC interface, and adds a rust goose based load test that works against the envoy proxy which should help simulate more real world end-to-end stats.

Adds a deployment option for the distributed store.

Note: this PR stacks on PR #350 so that we can easily build a local container using docker-compose with the right feature enabled depending on the deployment.


limitador-server/sandbox/docker-compose-limitador-memory.yaml
```diff
@@ -3,6 +3,9 @@ version: '3.8'
 services:
   limitador:
     image: ${LIMITADOR_IMAGE:-limitador-testing}
+    build:
```
eguzki, line 6: and if I do not want to build image and use an existing one instead?

limitador-server/sandbox/docker-compose-limitador-distributed.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - "8080"
+      - "8081"
+      - "5001"
+    ports:
+      - "18080:8080"
+      - "18081:8081"
+      - "15001:5001"
+    volumes:
+      - ./limits.yaml:/opt/kuadrant/limits/limits.yaml
+  redis:
```
eguzki, line 27: redis is needed?

limitador-server/sandbox/README.md
```diff
@@ -17,14 +17,39 @@ Check out `make help` for all the targets.
...
+| Redis Secured            | `make deploy-redis-tls` | Uses Redis with TLS and password protected to store counters                                                   |
+| Redis Cached             | `make deploy-redis-cached` | Uses Redis to store counters, with an in-memory cache                                                          |
+| Redis Otel Instrumented  | `make deploy-redis-otel` | Uses redis to store counters, [instrumented with opentelemetry](redis-otel/README.md)                          |
+| Disk                     | `make deploy-disk` | Uses disk to store counters                                                                                    |
+| Distributed | `make deploy-distributed` | Counters are held in Limitador (ephemeral) but replicated to other Limitador servers. |
+
+
+### Running Multi Node Distributed Deployments
+
+The `make deploy-distributed` target can be connected to other Limitador servers but requires you to set the `PEER_ID` and `PEER_URLS` environment variables when you run the target.
```
eguzki, line 33: have you considered running 3 instances (or N) of limitador on the same docker compose instance?

limitador-server/sandbox/docker-compose-limitador-distributed.yaml
```diff
@@ -0,0 +1,28 @@
...
+services:
+  limitador:
+    image: limitador-testing-all-features
+    build:
+      context: ../..
+      dockerfile: Dockerfile
+      args:
+        - CARGO_ARGS=--all-features
+    depends_on:
+      - envoy
```
eguzki, line 12: envoy is not being used, is it?

eguzki:
> Note: this PR stacks on PR https://github.com/Kuadrant/limitador/pull/350 so that we can easily build a local container using docker-compose with the right feature enabled depending on the deployment.

I would not make this PR depend on #350 . We might want to merge this and not the other one.

limitador-server/sandbox/docker-compose-limitador-memory.yaml
chirino: feel free to test it out.  by default it should first pull the image if It does not exist and use that.  If it can't pull it, then it builds it.

limitador-server/sandbox/docker-compose-limitador-distributed.yaml
chirino: your right, I'll remove that.

limitador-server/sandbox/README.md
chirino: I guess I could add more variables for the ports.. might get challenging exposing all the ports on the host right?

limitador-server/sandbox/docker-compose-limitador-distributed.yaml
chirino: All the other limitador compose files have this dependency.  But Yeah that seems backwards.

chirino:
> I would not make this PR depend on #350 . We might want to merge this and not the other one.

Can't build the docker image with the `--all-features` flag on a mac without #350.

limitador-server/sandbox/docker-compose-limitador-memory.yaml
eguzki: cool stuff. Great contribution.

with empty local docker image repo it builds a new image `limitador-testing`
```
> make deploy-in-memory
....
❯ docker compose -p sandbox ps
NAME                  IMAGE                           COMMAND                  SERVICE     CREATED              STATUS              PORTS
sandbox-envoy-1       envoyproxy/envoy:v1.20-latest   "/docker-entrypoint.…"   envoy       About a minute ago   Up About a minute   10000/tcp, 0.0.0.0:18000->80/tcp, :::18000->80/tcp, 0.0.0.0:18001->8001/tcp, :::18001->8001/tcp
sandbox-limitador-1   limitador-testing               "limitador-server --…"   limitador   About a minute ago   Up About a minute   0.0.0.0:18080->8080/tcp, :::18080->8080/tcp, 0.0.0.0:18081->8081/tcp, :::18081->8081/tcp
sandbox-upstream-1    kennethreitz/httpbin            "gunicorn -b 0.0.0.0…"   upstream    About a minute ago   Up About a minute   80/tcp
```

When I want to use `quay.io/kuadrant/limitador:latest`, I can easily run it.
```
> make deploy-in-memory LIMITADOR_IMAGE=quay.io/kuadrant/limitador:latest
....
❯  docker compose -p sandbox ps
NAME                  IMAGE                               COMMAND                  SERVICE     CREATED              STATUS              PORTS
sandbox-envoy-1       envoyproxy/envoy:v1.20-latest       "/docker-entrypoint.…"   envoy       About a minute ago   Up About a minute   10000/tcp, 0.0.0.0:18000->80/tcp, :::18000->80/tcp, 0.0.0.0:18001->8001/tcp, :::18001->8001/tcp
sandbox-limitador-1   quay.io/kuadrant/limitador:latest   "limitador-server --…"   limitador   About a minute ago   Up About a minute   0.0.0.0:18080->8080/tcp, :::18080->8080/tcp, 0.0.0.0:18081->8081/tcp, :::18081->8081/tcp
sandbox-upstream-1    kennethreitz/httpbin                "gunicorn -b 0.0.0.0…"   upstream    About a minute ago   Up About a minute   80/tcp
```

limitador-server/sandbox/README.md
eguzki: you do not need to worry about port conflict. Each limitador instance will be a unique container with dedicated assigned IP in the assigned docker network. Different IPs.. no port conflict. Docker magic.

limitador-server/sandbox/docker-compose-limitador-distributed.yaml
eguzki: yeah, exactly.

We are pushing the current "unique" sandbox too far. In other PR, we could refactor to have different sandboxes, each one running only necessary services. Each sandbox with custom readme.md and docker-compose.yaml file.

alexsnaps:
I'm somewhat confused about this pull request tbh... I'm probably missing something, but what's the intent here?

alexsnaps:
Just to be clear tho... not against it, just missing how/when I'd use this

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["limitador", "limitador-server"]
+members = ["limitador", "limitador-server", "limitador-server/sandbox/loadtest"]
```
alexsnaps, line 2: I wouldn't add this to the workspace here... There is no dependency and `loadtest` is really not useful in the context of limitador work, yet that'd impose build times et al...

chirino:
squashed and removed #350

limitador-server/sandbox/README.md
```diff
@@ -113,6 +146,24 @@ EOM
...
+```
+
+### Load Testing via Envoy Proxy
+
+```bash
+cargo run --manifest-path loadtest/Cargo.toml  --package loadtest --release -- --report-file=report.htm
+```
+
+The report will be saved in `report.htm` file.
+
```
chirino, line 166: @alexsnaps see above re: how to use

chirino:
@eguzki 3 node deployment option added.

limitador-server/sandbox/docker-compose-limitador-distributed.yaml
chirino: I fixed the dependency on this compose file and the other ones too.

limitador-server/sandbox/README.md
```diff
@@ -17,14 +17,40 @@ Check out `make help` for all the targets.
...
+| Disk                     | `make deploy-disk` | Uses disk to store counters                                                                                    |
+| Distributed | `make deploy-distributed` | Counters are held in Limitador (ephemeral) but replicated to other Limitador servers. |
+
+| Distributed 3 Node | `make deploy-distributed-3-node` | Counters are held in Limitador (ephemeral) but replicated to 3 other Limitador servers. |
+
+### Running Multi Node Distributed Deployments
+
+The `make deploy-distributed` target can be connected to other Limitador servers but requires you to set the `PEER_ID` and `PEER_URLS` environment variables when you run the target.
+
+If you have 3 servers you want to replicate between, you would run the following commands:
```
eguzki, line 36: no need to have 3 servers, right?

eguzki (commented):
Looking good.

left few comments.

After this PR is being merged, I probably will refactor a bit to have multiple "sandboxes" instead a "big" one. More scalable (easier to add new ones).

limitador-server/sandbox/README.md
chirino: Well, you might still want to do this on 3 machines if you are benchmarking.  Doing it all on one would not be very representative.

eguzki: I buy that

Dockerfile
```diff
@@ -4,20 +4,34 @@
 
 # Use bullseye as build image instead of Bookworm as ubi9 does not not have GLIBCXX_3.4.30
 # https://access.redhat.com/solutions/6969351
-FROM --platform=${BUILDPLATFORM} rust:1.78.0-bullseye as limitador-build
```
eguzki, line 7: is this breaking cross platform builds?

Dockerfile
```diff
@@ -4,20 +4,34 @@
...
-COPY . .
+# We set the env here just to make sure that the build is invalidated if the args change
+ENV CARGO_ARGS=${CARGO_ARGS}
 
-RUN cargo build --release
+# The following allows us to cache the Cargo dependency downloads with image layers
+COPY Cargo.toml Cargo.lock ./
+COPY limitador/Cargo.toml ./limitador/
+COPY limitador-server/Cargo.toml ./limitador-server/
+COPY limitador-server/sandbox/loadtest/Cargo.toml ./limitador-server/sandbox/loadtest/
```
eguzki, line 26: Maybe nitpicking here. Would it make sense a custom Dockerfile for the loadtesting tool?

Dockerfile
chirino: why would it? This Dockerfile works fine on my arm mac.

Dockerfile
chirino: dropping that line since the loadtest is not being built in docker.

alexsnaps (approved):

Dockerfile
eguzki: https://docs.docker.com/reference/dockerfile/#from

```
The optional --platform flag can be used to specify the platform of the image in case FROM references a multi-platform image. 
For example, linux/amd64, linux/arm64, or windows/amd64. By default, the target platform of the build request is used. Global build arguments can be used in the value of this flag, for example [automatic platform ARGs](https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope) 
allow you to force a stage to native build platform (--platform=$BUILDPLATFORM), and use it to cross-compile to the target platform inside the stage.
```

That was there for a reason and now you are removing it. Is there a reason for that or the optional `--platform=${BUILDPLATFORM}` was useless?

Dockerfile
```diff
@@ -4,20 +4,33 @@
...
+# We set the env here just to make sure that the build is invalidated if the args change
+ENV CARGO_ARGS=${CARGO_ARGS}
 
-RUN cargo build --release
+# The following allows us to cache the Cargo dependency downloads with image layers
+COPY Cargo.toml Cargo.lock ./
+COPY limitador/Cargo.toml ./limitador/
+COPY limitador-server/Cargo.toml ./limitador-server/
+RUN mkdir -p limitador-server/src && echo 'fn main() {}' > limitador-server/src/main.rs
+RUN mkdir -p limitador-server/sandbox/loadtest/src && echo 'fn main() {}' > limitador-server/sandbox/loadtest/src/main.rs
```
eguzki, line 27: is this still needed?

eguzki (changes requested):

Dockerfile
chirino: Yeah, I think it's best if the plain Dockerfile just does a container build that matches the host, and then we have additional Dockerfiles to do the cross compiles like [Dockerfile.aarch64](https://github.com/Kuadrant/limitador/blob/main/Dockerfile.aarch64).

Dockerfile
chirino: yep that's not needed either..

PR #352, opened 2024-06-04
First step: Limit storage

alexsnaps:
Slowly refactoring this, so the idea is that: 
 - [x] `Limit` definitions are now stored in `Arc`, so that they can be shared _within_ the crate's boundaries (i.e. towards the `(Async)CounterStorage`
 - [x] Do _not_ expose `Counter` as part of the lib's API, only as part of the SPI, i.e. `(Async)CounterStorage` 
 - [x] Have `Counter.limit` be an `Arc` coming from the `Storage.limit`'s

### DIFF
--- a/limitador/src/lib.rs
+++ b/limitador/src/lib.rs
@@ -226,6 +226,49 @@ pub struct CheckResult {
     pub limit_name: Option<String>,
 }
 
+impl CheckResult {
+    pub fn response_header(&mut self) -> HashMap<String, String> {
+        let mut headers = HashMap::new();
+        // sort by the limit remaining..
+        self.counters.sort_by(|a, b| {
+            let a_remaining = a.remaining().unwrap_or(a.max_value());
+            let b_remaining = b.remaining().unwrap_or(b.max_value());
+            a_remaining.cmp(&b_remaining)
+        });
+
+        let mut all_limits_text = String::with_capacity(20 * self.counters.len());
+        self.counters.iter_mut().for_each(|counter| {
+            all_limits_text.push_str(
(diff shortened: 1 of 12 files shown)


limitador/src/storage/mod.rs
```diff
@@ -284,8 +300,11 @@ pub trait AsyncCounterStorage: Sync + Send {
         delta: u64,
         load_counters: bool,
     ) -> Result<Authorization, StorageErr>;
-    async fn get_counters(&self, limits: HashSet<Limit>) -> Result<HashSet<Counter>, StorageErr>;
-    async fn delete_counters(&self, limits: HashSet<Limit>) -> Result<(), StorageErr>;
+    async fn get_counters(
+        &self,
+        limits: &HashSet<Arc<Limit>>,
+    ) -> Result<HashSet<Counter>, StorageErr>;
+    async fn delete_counters(&self, limits: &HashSet<Arc<Limit>>) -> Result<(), StorageErr>;
```
alexsnaps, line 307: See the comment above on the sync version of the trait.

limitador/src/lib.rs
```diff
@@ -226,6 +226,49 @@ pub struct CheckResult {
     pub limit_name: Option<String>,
 }
 
+impl CheckResult {
+    pub fn response_header(&mut self) -> HashMap<String, String> {
```
alexsnaps, line 231: Thinking this should return something other that `String` as keys to our `Map` here, but not a big deal (I think) for now as there is only one supported impl.

chirino: Thanks for refactoring this here.

chirino (approved):
LGTM

PR #356, opened 2024-06-12
Support providing an optional id to limits/counters

chirino:
* add a key_for_counters_v2 function that uses the id as the key if set, otherwise uses the previous key encoding strategy.
* updated the distributed store to use key_for_counters_v2.  Since we can’t decode a partial counter from id based keys, we now also keep in memory the Counter in a counter field of the limits map.

### DIFF
--- a/limitador/src/limit.rs
+++ b/limitador/src/limit.rs
@@ -51,6 +51,8 @@ impl From<String> for Namespace {
 
 #[derive(Eq, Debug, Clone, Serialize, Deserialize)]
 pub struct Limit {
+    #[serde(skip_serializing, default)]
+    id: Option<String>,
     namespace: Namespace,
     #[serde(skip_serializing, default)]
     max_value: u64,
@@ -319,6 +321,7 @@ impl Limit {
     {
         // the above where-clause is needed in order to call unwrap().
         Self {
+            id: None,
             namespace: namespace.into(),
             max_value,
             seconds,
@@ -333,6 +336,14 @@ impl Limit {
 
     pub fn namespace(&self) -> &Namespace {
         &self.namespace
+    }
+
+    pub fn set_id(&mut self, value: String) {
+        self.id = Some(value);
+    }
+
+    pub fn id(&self) -> &Option<String> {
+        &self.id
     }
 
     pub fn max_value(&self) -> u64 {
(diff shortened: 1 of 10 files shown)


limitador/src/storage/keys.rs
```diff
@@ -187,6 +221,20 @@ pub mod bin {
...
+        let mut encoded_key = Vec::new();
+        if counter.id().is_some() {
+            let key: IdCounterKey = counter.into();
+            encoded_key = postcard::to_extend(&1u8, encoded_key).unwrap();
+            encoded_key = postcard::to_extend(&key, encoded_key).unwrap();
+        } else {
+            let key: CounterKey = counter.into();
+            encoded_key = postcard::to_extend(&2u8, encoded_key).unwrap();
+            encoded_key = postcard::to_extend(&key, encoded_key).unwrap()
+        }
```
chirino, line 234: A hard choice was made here... we prefix the encoding with 0x01 or 0x02 so that a future decoder knows which type of encoding was used.  The problem is that this makes it incompatible with the existing `key_for_counter` encoding.    I guess we could commit the `encoded_key = postcard::to_extend(&2u8, encoded_key).unwrap();` line to remain compatible, but then the encoding is non-determinist.  You could maybe even figure out how to craft an id so that it's encoding matches an existing key in the `key_for_counter` format.

limitador/src/limit.rs
```diff
@@ -307,6 +311,7 @@ where
 
 impl Limit {
     pub fn new<N: Into<Namespace>, T: TryInto<Condition>>(
+        id: Option<String>,
```
alexsnaps, line 314: Note to mostly myself here, as I'll probably pick this up, but... 
Let's add a `Limit::with_id(String, ..)` or something, and keep `::new` as is, implying the `None`. That would be less of change across all the files, and make I think for a cleaner API

chirino: Good idea. ❤️ in flight wifi.

chirino:
moving the id out of the constuctor drastically reduced the size of the change set.

chirino:
after running the `rate_limited_id_counter_with_sync_redis` the keys in redis look like:

```
root@255340cd3402:/data# redis-cli --scan --pattern '*'
"\x02\x1ctest-rate_limited_id_counter"
"\x02\x1ctest-rate_limited_id_counter\x01\x06app_id\x0btest_app_id"
```

alexsnaps (approved):
Wondering... 
 - [ ] Should we push the version in our cargo.toml to `2.0.0-dev` (and `1.0.0-dev` for the crate)?
 - [ ] Should we add a test that actually verifies the encoded bytes for the new method, so to make sure they remain stable ?

chirino:
Since I'm new to rust, I'm not sure about the "Should we push the version in our cargo.toml to 2.0.0-dev (and 1.0.0-dev for the crate)?" bit.

PR #357, opened 2024-06-27
Bump async-trait to 0.1.80 to match Cargo.lock

jjkiely:
Hey folks :wave: 

I'm submitting this PR to bump the version of async-trait to match the version listed in [Cargo.lock](https://github.com/Kuadrant/limitador/blob/71ea11edf8eb1590ada1fba6c6a68f54429515f5/Cargo.lock#L327)

While these versions remain mismatched, offline installs are broken.


adam-cattermole (approved):
lgtm

adam-cattermole:
I am actually not sure how this will help but happy to merge if you think it's needed. The version in the toml is just the broad definition of the version required whereas the version in the Cargo.lock is the one that is actually being used when generated by cargo. So in this case "0.1" covers "0.1.80", and the version being used is "0.1.80". If we specify "0.1.80" and a new "0.1.81" is released we won't move to that one.

eguzki:
Not sure we should be doing this. For the limitador-server it makes  sense to lock down the versions (that is why a lock file exists). But limitador is a create (a lib being imported) and does not have a lock file. And as a library, dependencies should be more open (obviously with some restrictions). 

Reading https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

`async-trait = "0.1"` means that if `0.2'' is available, it will be installed, but never `1,0`.

If we change to `async-trait = "0.1.80"`, then when `"0.1.81"` or `0.1.90` is avaiable, that one will be used (if we want). However, it will never be updated to `0.2`. 

Not a big deal, TBH, but not sure this is what we want. I think we want to support anything v0.Y.Z. Because it should not break backwards compat and still get new fixes and features.

Anyway, I do not think that your issue would be gone, because if 0.1.81 is made available, we can upgrade the lock file to 0.1.81 and your issue will happen again. Am I right?

adam-cattermole:
@jjkiely and I managed to solve the offline building issue here and I think we're all in agreement that this change should not be merged so I'll close.

Issue #368, opened 2024-08-21
Investigate license of librocksdb-sys


david-martin:
See FOSSA report at https://app.fossa.com/projects/custom%2B162%2Fgit%2Bgithub.com%2FKuadrant%2Flimitador/refs/branch/main/e181a5aa3e6acbd3429923edd61b35a43d5abf17/issues/licensing?page=1&count=20&sort=issue_count_desc&grouping=revision&status=active&filter%5Btype%5D%5B0%5D=policy_flag&revisionScanId=62731053
which reports an AGPL-3.0 license.
The dependency has some specific licensing info at https://github.com/percona/PerconaFT?tab=readme-ov-file#license
The crate page lists 3 licenses (MIT, Apache2, BSD3) https://crates.io/crates/librocksdb-sys
However, the problem highlights is in some files like this:
<img width="761" alt="image" src="https://github.com/user-attachments/assets/548912d9-4c59-48a3-93bd-dc209ab612b4">

Options are to rule out AGPL as the other more favourable licenses are actually applicable to what's being pulled in from the dependency, or find an alternative library with a suitable license.

alexsnaps:
> Options are to rule out AGPL as the other more favourable licenses are actually applicable to what's being pulled in from the dependency, or find an alternative library with a suitable license.

I guess I'm confused as to what's "a suitable license"? Limitador (and all of Kuadrant afaik) is [Apache v2](https://github.com/Kuadrant/limitador/blob/main/LICENSE), this dependency is distributed under Apache v2 (along other licenses)... including that [very file](https://github.com/facebook/rocksdb/blob/main/utilities/transactions/lock/range/range_tree/lib/locktree/concurrent_tree.cc#L39-L49) that's referenced in the screenshot, that's indeed from another library, that again states [that part is Apache v2](https://github.com/percona/PerconaFT/?tab=readme-ov-file#license)... so I am really confused what's being discussed here? How AGPL would need to be an exception? Or what "a suitable license" would be in this case? Is the dual/triple licensing the issue? Am I missing some other files that are indeed not okay from that report?

david-martin:
@alexsnaps Let's close it.
After discussion at the time of reporting, the license issues reported here were 'Ignored' in FOSSA. I believe the reason for that decision was because of what you said above. 
![Image](https://github.com/user-attachments/assets/e8f14ae9-a8ac-4a2c-aed0-213d0758d2cd)

PR #369, opened 2024-08-21
gh action: license scan by fossa

eguzki:


eguzki:
Failing because we actually have an issue with the licenses. Explained in https://github.com/Kuadrant/limitador/issues/368

david-martin:
@eguzki the details are a little lacking in the logs :(
https://github.com/Kuadrant/limitador/actions/runs/10493245985/job/29066736342?pr=369
I wonder if that would give more info if a non push-only key was used?
Or if there's a way to provide a handy link to the fossa project page?

eguzki:
Changing the key no another one with full permissions.

After telling Fossa to ignore the dual license false positive, tests now pass.

eguzki:
Fossa scanning/testing works as follows: First step is scan phase. That scanning does not happen locally, but at Fossa servers. The report is then available in their site on completion.

If you check the scan step in our workflow, you will see the scan step provides a link: https://app.fossa.com/projects/custom%2b162%2fgit%2bgithub.com%2fKuadrant%2flimitador/refs/branch/fossa-scan/0458484c3719b3141af936912562731164ac24a4

The run step waits for scan completion and then fails if there are issues. 

PS: The link above no longer shows issues, because latest scan on the same revision returned no issues.

david-martin (approved):
Thanks for adding this @eguzki 
Good to have this automated at PR time

PR #373, opened 2024-09-16
Ignore unknown lints

alexsnaps:
This makes `-- -D warnings` feasible on older toolchains, so that the newer lint (unknown to older clippy) doesn't create a warning


didierofrivia (approved):
🧑‍🏭

PR #382, opened 2024-10-08
Update Redis crate and use pipeline to update counters

alexsnaps:


alexsnaps:
Sorry, added another [commit](https://github.com/Kuadrant/limitador/pull/382/commits/789d713a9ca52ac3b76ca30521510661bb38695e), that handles the absence of script on the server, just like [`Script::invoke_async`](https://github.com/redis-rs/redis-rs/blob/main/redis/src/script.rs#L166-L181) does, but which won't happen when `query_async` on a pipeline happens... The idea is to be safe on not needing to reload the scripts when reconnecting to a fresh/new server.

eguzki (approved):
LGTM

Issue #385, opened 2024-11-14
🔥 `limit::conditions` and have `Condition`s be proper CEL `Predicate`s


alexsnaps:
So to leverage CEL consistently across all Kuadrant components

 - [x] Parse `Limit`'s conditions as CEL expression (on creation)
 - [x] Parse `Limit`'s variables as CEL expression (again, on creation)
 - [x] 🔥 feature `lenient_conditions`
 - [x] 🔥 module `crate::limit::conditions`
 - [x] Inject context about the `Limit` itself in the interpreter to enable e.g. `Limit::id` lookup (one context for the whole `Limit`, i.e. all `condition`s and all `variable`s being resolved)
 - [ ] Should the descriptor(s) and their entries be first class citizens and have their own root binding?

PR #386, opened 2024-11-15
Proper `Error` types for the crate

### LINKED ISSUE #131: Have proper error for the crate
With trying to fix #100 in #130 - it turns out there isn't really a way to consistently report error with the underlying storage. 
We probably should come up with an `error` API to our public API that would let users consistently deal with error conditions being faced.

alexsnaps:
Fixes #131 

This is a first pass at this. `LimitadorError::InterpreterError` will "only" be used as a mean to cover for runtime errors, e.g. an expression that can't be resolved (see #385). Also, when doing that work, we should be able to return to a `Limit::new -> Self` over the current result and clean that API up a bit. 

Pretty sure other cases will emerge for `Limitador::Error` as we move towards [v2.0](https://github.com/Kuadrant/limitador/milestone/7) of the server, but hopefully `v1` for the crate 🤞


limitador/src/errors.rs
```diff
@@ -1,14 +1,51 @@
...
+    }
+}
+
+impl From<ConditionParsingError> for LimitadorError {
+    fn from(err: ConditionParsingError) -> Self {
+        LimitadorError::InterpreterError(err)
+    }
+}
+
+impl From<Infallible> for LimitadorError {
```
eguzki, line 47: nit: 
out of curiosity, why do we need to add this implementation?

When I apply this patch, it is not needed.


```diff
diff --git a/limitador/src/errors.rs b/limitador/src/errors.rs
index d590aaf..3de262e 100644
--- a/limitador/src/errors.rs
+++ b/limitador/src/errors.rs
@@ -1,6 +1,5 @@
 use crate::limit::ConditionParsingError;
 use crate::storage::StorageErr;
-use std::convert::Infallible;
 use std::error::Error;
 use std::fmt::{Display, Formatter};
 
@@ -43,9 +42,3 @@ impl From<ConditionParsingError> for LimitadorError {
         LimitadorError::InterpreterError(err)
     }
 }
-
-impl From<Infallible> for LimitadorError {
-    fn from(value: Infallible) -> Self {
-        unreachable!("unexpected infallible value: {:?}", value)
-    }
-}
diff --git a/limitador/src/limit.rs b/limitador/src/limit.rs
index 94b95b1..511f083 100644
--- a/limitador/src/limit.rs
+++ b/limitador/src/limit.rs
@@ -1080,7 +1080,7 @@ mod tests {
             limit1.namespace.clone(),
             limit1.max_value + 10,
             limit1.seconds,
-            limit1.conditions.clone(),
+            vec!["req.method == 'GET'"],
             limit1.variables.clone(),
         )
         .expect("This must be a valid limit!");
```

Not arguing to change it, just curiosity.

PR #387, opened 2024-11-20
e2e limits_file_watcher_on_docker

eguzki:


limitador-server/sandbox/docker-compose-limitador-memory.yaml
```diff
@@ -27,4 +26,6 @@ services:
       - "18080:8080"
       - "18081:8081"
     volumes:
-      - ./limits.yaml:/opt/kuadrant/limits/limits.yaml
+      # the entire path needs to be mounted
+      # or inotify events are not triggered for updates on the mounted file
```
alexsnaps, line 30: This part?

eguzki:
> What is this test trying to assert? That we can observe filesystem changes when running within a docker container?

Yes, 
Limitador has some coding regarding watching files in dockerized envs https://github.com/Kuadrant/limitador/blob/main/limitador-server/src/main.rs#L268-L289

This e2e test makes sure it works and nothing broke it.

limitador-server/sandbox/docker-compose-limitador-memory.yaml
eguzki: This is a comment to warn about some configuration that, for unknown reason to me yet, events are not triggered when the file changes. That happens exactly when the file is mounted. 

When the parent dir is mounted, then events are triggered whenever the limits file is updated.

alexsnaps (approved):
Right, but that's the same behavior as if watching the local FS... So not saying this isn't useful, but it's a lot of jumping through hoops to validate this... 
The one downside I see, is that we keep on adding things to the e2e w/o adding much safety nets at the ~unit~ `cargo test` layer. I personally favor quick feedback loops, and e2e tests are both cumbersome and slow to run (and generally will run on CI, which makes that time even longer). 
I guess lets add this, see it ever catches anything and see from there. i.e. if it fails, lets make sure we add some quick test that prove its fixed, so we catch it sooner in the future.

PR #389, opened 2024-11-21
Proper `impl Ord for Limit`

alexsnaps:
`impl Ord for Limit` was out of sync with `Eq`... So looking up counters would fail when navigating the `BTreeMap` 🤦 
My bad!

### DIFF
--- a/limitador/src/limit.rs
+++ b/limitador/src/limit.rs
@@ -49,7 +49,7 @@ impl From<String> for Namespace {
     }
 }
 
-#[derive(Eq, Debug, Clone, PartialOrd, Ord, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


limitador/src/limit.rs
```diff
@@ -49,7 +49,7 @@ impl From<String> for Namespace {
     }
 }
 
-#[derive(Eq, Debug, Clone, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Eq, Debug, Clone, Serialize, Deserialize)]
```
eguzki, line 52: What was the default `PartialOrd` and `Ord` doing instead?

eguzki (approved):

Issue #62, opened 2022-05-12
all the descriptors are consolidated in a single descriptor


eguzki:
For the following Envoy's RLS request:

```yaml
domain: example.org    
descriptors:           
  - entries:           
    - C: VALUE         
  - entries:           
    - A: VALUE         
    - B: VALUE         
```

the following ratelimit config would be activated and it should not.
```yaml
conditions: []  
max_value: 1                      
seconds: 60                       
variables: ["A", "B", "C"]             
namespace: example.org            
```

alexsnaps:
@didierofrivia may be that could be interesting for you to look into…

alexsnaps:
See #354 ... maybe we should consider having `descriptors` as a root binding and have it as a `List` of `Map<String, String>`... ?

PR #396, opened 2024-11-26
🔥 conditions et at, replacing with proper `cel::Predicate`s & `::Expression`s

### LINKED ISSUE #385: 🔥 `limit::conditions` and have `Condition`s be proper CEL `Predicate`s

alexsnaps:
fix #385

### DIFF
--- a/limitador/src/limit/cel.rs
+++ b/limitador/src/limit/cel.rs
@@ -1,0 +1,162 @@
+use crate::limit::cel::errors::EvaluationError;
+use cel_interpreter::{ExecutionError, Value};
+pub use errors::ParseError;
+
+pub(super) mod errors {
+    use cel_interpreter::ExecutionError;
+    use std::error::Error;
+    use std::fmt::{Display, Formatter};
+
+    #[derive(Debug, PartialEq)]
+    pub enum EvaluationError {
+        UnexpectedValueType(String),
+        ExecutionError(ExecutionError),
+    }
+
+    impl Display for EvaluationError {
+        fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+            match self {
+                EvaluationError::UnexpectedValueType(value) => {
+                    write!(f, "unexpected value of type {}", value)
+                }
+                EvaluationError::ExecutionError(error) => error.fmt(f),
+            }
+        }
+    }
+
+    impl Error for EvaluationError {}
+
+    #[derive(Debug)]
+    pub struct ParseError {
+        input: String,
+        source: Box<dyn Error + 'static>,
+    }
+
+    impl ParseError {
+        pub fn from(source: cel_parser::ParseError, input: String) -> Self {
+            Self {
+                input,
+                source: Box::new(source),
+            }
+        }
+    }
+
+    impl Display for ParseError {
+        fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+            write!(f, "couldn't parse {}: {}", self.input, self.source)
+        }
+    }
+
+    impl Error for ParseError {
+        fn source(&self) -> Option<&(dyn Error + 'static)> {
+            Some(self.source.as_ref())
+        }
+    }
+
+    impl From<ExecutionError> for EvaluationError {
+        fn from(err: ExecutionError) -> Self {
+            EvaluationError::ExecutionError(err)
+        }
+    }
+}
+
+pub struct Context {}
+
+pub struct Expression {
+    expression: cel_parser::Expression,
+}
+
+impl Expression {
+    pub fn parse(source: &str) -> Result<Self, ParseError> {
+        match cel_parser::parse(source) {
+            Ok(expression) => Ok(Self { expression }),
+            Err(err) => Err(ParseError::from(err, source.to_string())),
+        }
+    }
+
+    pub fn eval(&self, ctx: &Context) -> Result<String, EvaluationError> {
+        match self.resolve(ctx)? {
+            Value::Int(i) => Ok(i.to_string()),
+            Value::UInt(i) => Ok(i.to_string()),
+            Value::Float(f) => Ok(f.to_string()),
+            Value::String(s) => Ok(s.to_string()),
+            Value::Null => Ok("null".to_owned()),
+            Value::Bool(b) => Ok(b.to_string()),
+            val => Err(err_on_value(val)),
+        }
+    }
+
+    pub fn resolve(&self, _ctx: &Context) -> Result<Value, ExecutionError> {
+        let ctx = cel_interpreter::Context::default();
+        Value::resolve(&self.expression, &ctx)
+    }
+}
+
+fn err_on_value(val: Value) -> EvaluationError {
+    match val {
+        Value::List(list) => EvaluationError::UnexpectedValueType(format!("list: `{:?}`", *list)),
+        Value::Map(map) => EvaluationError::UnexpectedValueType(format!("map: `{:?}`", *map.map)),
+        Value::Function(ident, _) => {
+            EvaluationError::UnexpectedValueType(format!("function: `{}`", *ident))
+        }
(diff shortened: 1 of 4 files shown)


limitador/src/counter.rs
```diff
@@ -120,3 +139,28 @@ impl PartialEq for Counter {
...
+            Vec::default(),
+            ["int(x) * 3".try_into().expect("failed parsing!")],
+        );
+        let key = "x".to_string();
+        let counter = Counter::new(limit, HashMap::from([(key.clone(), "14".to_string())]))
+            .expect("failed creating counter");
+        assert_eq!(
+            counter.set_variables.get(&key),
+            Some("42".to_string()).as_ref()
+        );
```
alexsnaps, line 164: So I guess this is now new...

limitador/src/limit/cel.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+}
+
+pub struct Context<'a> {
+    variables: HashSet<String>,
+    ctx: cel_interpreter::Context<'a>,
+}
+
+impl Context<'_> {
+    pub(crate) fn new(limit: &Limit, root: String, values: &HashMap<String, String>) -> Self {
```
alexsnaps, line 82: So, this `root` relates to this point from #385

>Should the descriptor(s) and their entries be first class citizens and have their own root binding?

Right now, `limit` binds the limit info (`id` & `name`), I was thinking that possibly we could have some root binding for the descriptor entries (e.g. `descriptor`), which... could then also be a `List` and hold multiple of them and all their entries: `descriptor[0].foo == '1' && descriptor[1].bar != 'bad'`... See [this issue here](https://github.com/Kuadrant/limitador/issues/62), so it'd probably be a `List<HashMap<String, String>>` as well then...

Whatever the case, who would set that `root` namespace/binding to something? I'd expect it to be the server code... but that then creates further issues in the (de)serialization paths... Also, the logic of resolving the `Limit.variables()` would need to account for that (along side what happens in the wasm-shim's code to lazily resolve attributes from the host).

alexsnaps:
Pretty big PR, but I tried to keep a somewhat meaningful history and broke it down according to the #385 
Thanks... and... sorry 🤦

limitador/src/storage/distributed/mod.rs
```diff
@@ -333,6 +334,15 @@ fn encode_counter_to_key(counter: &Counter) -> Vec<u8> {
 }
 
 fn encode_limit_to_key(limit: &Limit) -> Vec<u8> {
-    let counter = Counter::new(limit.clone(), HashMap::default());
+    // fixme this is broken!
```
alexsnaps, line 337: This has always been broken... and will need revisiting! But we can't just fake a the variables values like this (and they were simply left empty before, but this won't work now, as they'd fail resolution. I could "just" use the new `Counter::resolved_vars`, but that would just hide the problem a little more again.

limitador/src/limit/cel.rs
alexsnaps: Also, doing so would tie releasing this with a release of the kuadrant-operator... right now, this isn't necessary.

limitador/src/counter.rs
```diff
@@ -120,3 +139,31 @@ impl PartialEq for Counter {
...
+        let counter = Counter::new(
+            limit,
+            HashMap::from([("ts".to_string(), "2019-10-12T13:20:50.52Z".to_string())]),
+        )
+        .expect("failed creating counter");
+        assert_eq!(
+            counter.set_variables.get(var),
+            Some("13".to_string()).as_ref()
+        );
```
alexsnaps, line 167: So I guess this is now new...

limitador-server/examples/envoy.yaml
```diff
@@ -29,7 +29,7 @@ static_resources:
                     - stage: 0
                       actions:
                         - {request_headers: {header_name: "userid", descriptor_key: "user_id"}}
-                        - {request_headers: {header_name: ":method", descriptor_key: "req.method"}}
+                        - { request_headers: { header_name: ":method", descriptor_key: "req_method" } }
```
eguzki, line 32: What if, as a user, I want to send descriptor entry key `req.method`? 

If _Limit_'s _Variable_ and _Condition_ are full CEL expressions, there should be a way to make reference to the descriptor entries. Maybe something like:

```
req.entry["req.method"] == 'POST'
```

wdyt?

alexsnaps: That's indeed another reason for the ["root" binding I'm arguing for](https://github.com/Kuadrant/limitador/pull/396#discussion_r1861201539)... 
so `req` doesn't mean anything... right now, we use the descriptor entries to infer some naming... We _could_ do some massaging when getting the values inject still too... e.g. treat the dot (`.`) as some special thing and create a map... e.g. when getting `req.method` as a descriptor key, it'd bind to a map `req` with an entry `method`, which would point to the value of the descriptor... but that's a never ending story... what about spaces (` `), question marks (`?`), colons (`:`), ... ? More so, what if there is an entry `limit` in one of the descriptors?

So I'd "just" embrace the fact that they can be any character and put them all in a map... but what? Each and every discreet descriptor, i.e. `descriptors[0]["req.method:foo?bar"]` ?

alexsnaps: Also, to be clear, the `.` navigation works transparently on `Map`:

```
descriptor[0].foo == descriptor[0]['foo']
```

limitador/src/limit/cel.rs
alexsnaps: So... thinking about this a little more, and given [this comment](https://github.com/Kuadrant/limitador/pull/396#discussion_r1865738487), here's how I'm proposing to go about this:

- [ ] Have a blanket `From<HashMap<String, String>` for `Context`... which would _not_ add the binding for `limit`.
- [ ] Make `Context::new` public to all users
- [ ] Change `Limit::applies` & `::resolve_variables` to take a `&Context` instead of creating it explicitly
- [ ] Have the `(Async)RateLimiter::counters_that_apply` also take the `&Context`
- [ ] Have the `(Async)RateLimiter` APIs expose the `Context`, instead of `&HashMap<String, String>` 

That way, the server code can decide how it wires the `RateLimitDescriptor`s and their `.entries` from envoy's protocol for external rate limiting.

alexsnaps: > Have the (Async)RateLimiter::counters_that_apply also take the &Context
> Have the (Async)RateLimiter APIs expose the Context, instead of &HashMap<String, String>

duh! this obviously doesn't work 🤦 Or at least not while injecting the `Limit` this operates on.

alexsnaps:
As per [this](https://github.com/Kuadrant/limitador/pull/396#discussion_r1865856311), starting to address these issues

alexsnaps:
This now fixes #62 as well

adam-cattermole (approved):
After our discussion the other day and time to go through it, the changes look sound to me, so I'm happy to approve to unblock

eguzki:
Can we tell that this is a breaking change? Just to confirm the `breaking-change` I just added.

eguzki:
Is there any plan to have some documentation here or somewhere else? So, whenever I want to write some limitador configuration, what context do I have available in CEL expressions? How do I make reference to descriptor entries to write predicates?

eguzki:
After merging this, is it correct to say that Kuadrant's RatelimitPolicies are broken on _HEAD_ of `main` ?

PR #399, opened 2024-12-09
Delete counters of qualified limits

alexsnaps:
I don't think this is really a desirable fix to be honest... `O(n)` isn't really sustainable. 
Probably resorting to a ordered collection and have all qualified counters for a qualified limit next to each other would be better... I'll see what I can do as part of #397

### DIFF
--- a/limitador/src/storage/in_memory.rs
+++ b/limitador/src/storage/in_memory.rs
@@ -230,7 +230,15 @@ impl InMemoryStorage {
     }
 
     fn delete_counters_of_limit(&self, limit: &Limit) {
-        self.simple_limits.write().unwrap().remove(limit);
+        if limit.variables().is_empty() {
+            self.simple_limits.write().unwrap().remove(limit);
+        } else {
+            for (c, _) in self.qualified_counters.iter() {
+                if c.limit() == limit {
+                    self.qualified_counters.invalidate(&c);
+                }
+            }
+        }
     }
 
     fn counter_is_within_limits(counter: &Counter, current_val: Option<&u64>, delta: u64) -> bool {
--- a/limitador/src/lib.rs
+++ b/limitador/src/lib.rs
@@ -697,6 +697,7 @@ fn classify_limits_by_namespace(
 mod test {
     use crate::limit::{Context, Expression, Limit};
     use crate::RateLimiter;
+    use std::collections::HashMap;
 
     #[test]
     fn properly_updates_existing_limits() {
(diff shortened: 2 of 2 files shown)


adam-cattermole (approved):
Sorry for the slow review, I looked at it before and meant to get back to it

limitador/src/storage/in_memory.rs
```diff
@@ -230,7 +233,24 @@ impl InMemoryStorage {
     }
 
     fn delete_counters_of_limit(&self, limit: &Limit) {
-        self.simple_limits.write().unwrap().remove(limit);
+        if limit.variables().is_empty() {
+            self.simple_limits.write().unwrap().remove(limit);
+        } else {
+            let l = limit.clone();
+            match self
```
andrewdavidmackenzie, line 240: Do you run clippy?

It would suggest to me I use `if Let Err() = ...` as doing nothing in the Ok(_) branch.

alexsnaps: We do use it... is that a lint now tho? Can you link me to it? 
This uses both 1.75 and stable, and... I never got that.

alexsnaps: ... and I don't know how much I like it. I know I really hate the opposite (`if let Ok(...)`), but maybe less so for `Err`...

andrewdavidmackenzie:
```
    Checking pigg v0.6.0 (/Users/andrew/workspace/pigg)
warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
   --> src/hardware_subscription.rs:106:9
    |
106 | /         match gui_sender
107 | |             .send(SubscriptionEvent::Ready(subscriber_sender.clone()))
108 | |             .await
109 | |         {
...   |
113 | |             Ok(_) => {}
114 | |         }
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_match
    = note: `#[warn(clippy::single_match)]` on by default
help: try
    |
106 ~         if let Err(e) = gui_sender
107 +             .send(SubscriptionEvent::Ready(subscriber_sender.clone()))
108 +             .await {
109 +             report_error(&mut gui_sender, &format!("Send error: {e}")).await;
110 +         }
    |
```

alexsnaps: Ah, yeah `single_match`... which this sorta is...

PR #405, opened 2024-12-11
FOSSA license scan runs only on main repo

### LINKED ISSUE #402: Fix GH actions to work on forks!
Our actions are currently broken. I'm unsure to what extent, but at least the FOSSA one _still_ is, despite me trying to get it working. 
See [here](https://github.com/Kuadrant/limitador/actions/runs/12257886251) for the latest type of failure. I reverted that, but now that means that again, on forks and PR of forks, that step will be broken!

I don't know whether its the action itself, or what it is... but this whole `secrets` stuff is broken beyond belief! It was a mistake to put all this config stuff in the repo alongside the code, blame GH... but what can we do? Should we move all our CI out of this mess? I've searched, but can't seem to find what are the proper patterns to support this fork contribution workflow, which... is actually how GH is supposed to work, but... not GH actions? What am I missing?

eguzki:
Fixes #402

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -100,23 +100,19 @@ jobs:
             # set ref name tag for non-main branches
             type=raw,value=${{ github.ref_name }},enable=${{ github.ref_name != env.MAIN_BRANCH_NAME }}
       - name: Login to container registry
-        if: github.repository == 'Kuadrant/limitador'
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.IMG_REGISTRY_USERNAME }}
           password: ${{ secrets.IMG_REGISTRY_TOKEN }}
           registry: ${{ env.IMG_REGISTRY_HOST }}
       - name: Create manifest list and push
-        if: github.repository == 'Kuadrant/limitador'
         working-directory: /tmp/digests
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build-image.yaml
```diff
@@ -100,23 +100,19 @@ jobs:
             # set ref name tag for non-main branches
             type=raw,value=${{ github.ref_name }},enable=${{ github.ref_name != env.MAIN_BRANCH_NAME }}
       - name: Login to container registry
```
alexsnaps, line 102: Wait, we do want to login, ... _always_?

.github/workflows/release.yaml
```diff
@@ -12,7 +12,6 @@ on:
 jobs:
   publish:
     name: Publish Limitador crate to crates.io
```
alexsnaps, line 14: Why we would allow this on forks?

.github/workflows/license-scan.yaml
```diff
@@ -12,7 +12,7 @@ on:
 jobs:
   fossa-scan:
     name: Find license compliance and security issues
-    if: github.repository == 'Kuadrant/limitador'
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
alexsnaps, line 15: Hope this fixes it 🤞 
I had this job fail here once on `main`, but let's try this 🙌

alexsnaps (approved):
A little confused why some things were re-enabled back, but if it works its all I care about 😄

PR #412, opened 2025-01-30
/status endpoint adds status info

alexsnaps:
Added are the timestamp for server started and last successful limits file loaded. It also monotonically increases a config version on successful loading and applying of the file. Lastly, it counts how many time it failed to load a limits file since the last success.

## validating

1. start the server:
`./target/debug/limitador-server limitador-server/examples/limits.yaml -vvv`
2. hit the /status endpoint: `http://localhost:8080/status`
3. edit `limitador-server/examples/limits.yaml`
4. /status endpoint should have updated both `config_applied_at` and `config_version`
5. edit `limitador-server/examples/limits.yaml` with an invalid change
6. /status endpoint should have updated the `config_err_since` field


Cargo.toml
```diff
@@ -5,3 +5,4 @@ resolver = "2"
 [profile.release]
 lto = true
 codegen-units = 1
+strip = "symbols"
```
andrewdavidmackenzie, line 8: The docs (https://doc.rust-lang.org/cargo/reference/profiles.html) list the options, but don't really explain the difference between "debuginfo" and "symbols" :-( 
So, I had to go here: https://users.rust-lang.org/t/difference-between-strip-symbols-and-strip-debuginfo/96602 and learn it...thanks!

limitador-server/Cargo.toml
```diff
@@ -48,6 +48,7 @@ sysinfo = "0.32"
 openssl = { version = "0.10.66", features = ["vendored"] }
 metrics = "0.22.3"
 metrics-exporter-prometheus = "0.14.0"
+chrono = { version = "0.4", features = ["serde"] }
```
andrewdavidmackenzie, line 51: If you are looking to reduce binary size further (based on strip above) you could also look at:
- lto = fat
- opt-level = s (debateable, would need benchmarking on performance)
- codegen-units=1 (this helped reduce my binary size even further)
- use default-features=false in all your dependencies, then add back in the features you need. A bit laborious, but a one-time job.

limitador-server/src/http_api/server.rs
```diff
@@ -32,6 +41,10 @@ impl RateLimitData {
     fn metrics(&self) -> &PrometheusMetrics {
         self.metrics.as_ref()
     }
+
+    fn status(&self) -> Status {
+        *self.status.read().unwrap()
```
andrewdavidmackenzie, line 46: What's your policy on unwraps and possible panics?

limitador-server/src/http_api/server.rs
```diff
@@ -32,6 +41,10 @@ impl RateLimitData {
     fn metrics(&self) -> &PrometheusMetrics {
         self.metrics.as_ref()
     }
+
+    fn status(&self) -> Status {
```
andrewdavidmackenzie, line 45: Do you want to deref?
You could return &Status, like the others, and faster?

limitador-server/src/http_api/server.rs
```diff
@@ -61,8 +74,8 @@ impl ResponseError for ErrorResponse {
 
 // Used for health checks
 #[api_v2_operation]
-async fn status() -> web::Json<()> {
-    Json(())
+async fn status(data: web::Data<RateLimitData>) -> web::Json<Status> {
+    Json(data.get_ref().status())
```
andrewdavidmackenzie, line 78: I am assuming Json() can handle the parameter being &Status also.

limitador-server/src/main.rs
```diff
@@ -276,10 +312,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
                         if location == last_known_canonical_path {
                             let limiter = limiter.clone();
+                            let status_updater = Arc::clone(&status_updater);
+
                             handle.spawn(async move {
                                 match limiter.load_limits_from_file(&location).await {
-                                    Ok(_) => info!("data modified; reloaded limit file"),
-                                    Err(e) => error!("Failed reloading limit file: {}", e),
+                                    Ok(_) => {
+                                        status_updater.write().unwrap().config_success();
```
andrewdavidmackenzie, line 320: unwrap() here and below, not sure on the probability of status_updater.write() failing....
ah, not sure I follow why using write...but maybe me.

limitador-server/src/main.rs
```diff
@@ -258,6 +291,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // structure needed to keep state of the last known canonical limits file path
     let mut last_known_canonical_path = fs::canonicalize(&limit_file).unwrap();
 
+    let status: Arc<RwLock<Status>> = Arc::default();
+    let status_updater = Arc::clone(&status);
```
andrewdavidmackenzie, line 295: could you just use `status.clone()` ?
Not sure I follow this one...
Could create with Status::default() ?

andrewdavidmackenzie (commented):
Not sure I made any useful suggestions, a bit of a learning being my first read of this code. Hope it helps...

limitador-server/src/http_api/server.rs
alexsnaps: No you can't cause that's guarded by a `RwLock`. The reference can't outlive the `RwLockReadGuard`.
So instead of reading the fields, which are just a few bytes, and return that, I'd opted for "snapshot'ing" the state in a `Copy`.

limitador-server/src/http_api/server.rs
alexsnaps: Possibly, but see [above](https://github.com/Kuadrant/limitador/pull/412#discussion_r1935550205).

limitador-server/src/main.rs
alexsnaps: That's stylistic. They are the same. Usually the `Arc::clone` is preferred as it clearly indicate we are bumping a counter and not cloning the deref struct... in this case tho, yeah... it's obvious that `status` is an `Arc`

Cargo.toml
alexsnaps: I actually didn't mean to commit that change 🤦 
Let me undo that!

limitador-server/src/http_api/server.rs
alexsnaps: See [below](https://github.com/Kuadrant/limitador/pull/412#discussion_r1935555048)

limitador-server/src/main.rs
alexsnaps: Oh, and the `.write()` is because [`Status.config_success`](https://github.com/Kuadrant/limitador/pull/412/files#diff-77c4c3afe1e75f3f1602389f5895694a70395b1a7450ad7776945631b4bd28e9R223) requires a `&mut self`, as it mutates the struct. So we acquire the write lock of the `RwLock`

alexsnaps:
> Not sure I made any useful suggestions, a bit of a learning being my first read of this code. Hope it helps...

I realize that without much context, this might be somewhat weird indeed. So what's going on here, is that we have an "observer" on the the `limits.yaml` file that contains all `Limit` definition that will invoke a function when a change happens to that file. So that `Status` is stored on the heap, with multiple threads & async functions having a reference to it. The async function that reads the data is the `/status` resolver. While on every change to the file (on the file system), we update that data. 

I went for a `RwLock` over a `Mutex` to avoid serializing access when e.g. a bad client would hammer `/status` for a while, so that these can get served asap, and avoid waiting for too long on grabbing the write lock, should a change have occured. Now it all gets loaded and applied anyways _before_ we (wait for?) grabbing the write lock. 

The idea is that this `/status` now gives some insights about what "version" of the config is applied. As this can take some time when we do it through Kuadrant (and a mounted `ConfigMap`)

limitador-server/src/main.rs
alexsnaps: Ugh... and yeah, that uses `Status::default()` actually, only implicitly. This `Arc::default()`, calls into `RwLock::default()`, which calls into `Status::default()` because of the explicit type on the lhs (`: Arc<RwLock<Status>>`). So we get a reference counted read/write lock protecting our `Status`. Syntactic sugar ftw :) ... not so for readability maybe?

didierofrivia (approved):
Thanks for the clarification in the comments, overall looks good.. I reckon with the number of failures the user could identify the version of the limit config it's trying out with the respective changes, better than a simple boolean.

PR #418, opened 2025-03-14
fix: remove Cargo.lock from loadtest, confuses dependabot

alexsnaps:


alexsnaps:
>I wonder if instead it's better to tell dependabot to ignore this location but I guess we can add this back later if we want

Had a quick search, but didn't find it in the UIs I visited... now, honestly, there is no value to checking that `Cargo.lock` file this deleted I think

PR #424, opened 2025-06-17
Allow req.hits_addend value to be zero

eguzki:
Fixes https://github.com/Kuadrant/limitador/issues/425

### DIFF
--- a/limitador/src/storage/distributed/grpc/mod.rs
+++ b/limitador/src/storage/distributed/grpc/mod.rs
@@ -389,6 +389,7 @@ impl MessageSender {
             },
         }
     }
+
     fn try_reserve(&self) -> Result<MessagePermit<'_>, Status> {
         match self {
             MessageSender::Client(sender) => {
--- a/limitador/src/storage/distributed/mod.rs
+++ b/limitador/src/storage/distributed/mod.rs
@@ -17,6 +17,7 @@ use crate::storage::keys::bin::key_for_counter_v2;
 use crate::storage::{Authorization, CounterStorage, StorageErr};
 
 mod cr_counter_value;
+#[allow(clippy::result_large_err)]
 mod grpc;
 
 pub type LimitsMap = HashMap<Vec<u8>, Arc<CounterEntry>>;


eguzki:
Can you help fixing the clippy :dog: complaining @alexsnaps ?

limitador-server/src/envoy_rls/server.rs
```diff
@@ -562,9 +556,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_0_hits_addend_is_converted_to_1() {
-        // "hits_addend" is optional according to the spec, and should default
-        // to 1, However, with the autogenerated structs it defaults to 0.
```
alexsnaps, line 567: Well... that's a problem! Now it'd be fine _for us always_, as we do explicitly set it tho.

eguzki:
Thanks

alexsnaps (approved):
So, that comment seemed wrong about `hits_addend` being "optional", if [I read this correctly](https://github.com/envoyproxy/envoy/blob/f5280097c3c7b1a6fd6c09ff4e8adcefb8540a8e/api/envoy/service/ratelimit/v3/rls.proto#L54)

alexsnaps:
> Thanks

Should be good now, sorry for the force push. Added a blanket `allow` on the module, that `Status` is used all over the place as a `Result::Err`

limitador-server/src/envoy_rls/server.rs
alexsnaps: Well the proto seems to say it's not optional (or is it because it's the last field?!), so it has to be set, hasn't it?!

alexsnaps:
> So, that comment seemed wrong about `hits_addend` being "optional", if [I read this correctly](https://github.com/envoyproxy/envoy/blob/f5280097c3c7b1a6fd6c09ff4e8adcefb8540a8e/api/envoy/service/ratelimit/v3/rls.proto#L54)

👆 @eguzki

alexsnaps (approved):

PR #430, opened 2025-06-20
kuadrant gRPC service: check rate limits

### LINKED ISSUE #429: Read-only operation to check rate limits
Currently, the Limitador RLS server implements Envoy's [Rate Limit Service (RLS)](https://github.com/envoyproxy/envoy/blob/bb803065c25affa40d89f3184798c0778276b940/api/envoy/service/ratelimit/v3/rls.proto#L36):

```proto
service RateLimitService {
  // Determine whether rate limiting should take place.
  rpc ShouldRateLimit(RateLimitRequest) returns (RateLimitResponse) {
  }
}
```

This service relies on the RateLimitRequest object, which includes a field called `hits_addend`, representing the increment value as a positive integer. Although this field may not be explicitly defined in the request, the specification clearly states that its default value is 1. Additionally, it's not an optional field, meaning there's no way to indicate that it was intentionally set to 0 or left unset. Therefore, if the server receives a value of 0, it must treat it as 1.

As a result, the ShouldRateLimit method always mutates state—it is effectively a write-only operation.

Read-only operation to check rate limits behavior is necessary to support the TokenRateLimitPolicy ([issue](https://github.com/Kuadrant/kuadrant-operator/issues/1330), [RFC](https://github.com/Kuadrant/architecture/pull/120)). In summary, Kuadrant checks the remaining token budget during the request phase and updates the token usage during the response phase.

eguzki:
### What

Fixes #429


eguzki:
Some items to be discussed before moving on to "ready for review"
### Metrics
Should we add custom metrics for the new gRPC service method? Currently we have (relevant ones):
````
# HELP authorized_calls Authorized calls
# TYPE authorized_calls counter
authorized_calls{limitador_namespace="test_namespace"} 19

# HELP limited_calls Limited calls
# TYPE limited_calls counter
limited_calls{limitador_namespace="test_namespace"} 1
````
### Counters in response
Should we add the [rate limit headers](https://datatracker.ietf.org/doc/id/draft-polli-ratelimit-headers-03.html) on the response of the new service method? It should be fairly easy to do. I tend to think we should. 

### Tracing headers

Should we implement the propagation of the tracing headers?

eguzki:
clippy complaining now about `protoc failed: udpa/annotations/status.proto: File not found` and it did not before. Maybe for some unknown reason https://github.com/Kuadrant/limitador/pull/426 is required now?

Patryk-Stefanski:
code looks good and +1 on adding counters in response. As for metrics, I'm not sure if we need anything beyond what we have at this point. We can revisit this maybe when we're implementing some observability for AI gateways.

PR #433, opened 2025-07-21
Ensures the value does not exceed a counter's max_value

eguzki:


eguzki:
the code assumes 'value' will not exceed 'max_value'. If it does, the unsigned
 subtraction to compute `remaining` will underflow, making 'remaining' a large positive number.

alexsnaps:
> the code assumes 'value' will not exceed 'max_value'. If it does, the unsigned subtraction to compute `remaining` will underflow, making 'remaining' a large positive number.
> 
> Does it make sense?

It does, but [just deal with the overflow](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4a00da2aed8815e3ce8f04fa87c34cca)... confused. Which is what you're sorta doing, but on "the other side", which isn't thread safe.

PR #442, opened 2025-08-28
Metric labels

alexsnaps:
This adds `-L <labels.yaml>` to use custom labels in metrics. `labels.yaml` is a `Map<String, String>`, where the values `String` will be converted to CEL `Expression`, any failure is ignored (I don't think metrics should be stopping the server or otherwise be a bigger problem). 

These CEL expressions should probably be `check` 'ed for way before they even make it to the `ConfigMap` that will configure this.

Interestingly, ` -l, --limit-name-in-labels` can also now be a `labels.yaml` file with a `limit_name` (or whatever the user actually wants) entry:

```yaml
limit_name: limit.name
other_label: descriptors[0].some_descriptor_entry
```

we could also [expose `limit.namespace`](https://github.com/Kuadrant/limitador/blob/fa416b9227e6205ed5e6ef8a2f2944d1f770fba9/limitador/src/limit/cel.rs#L111-L139) and then get rid of any hard coded labels maybe?

### DIFF
--- a/limitador-server/src/prometheus_metrics.rs
+++ b/limitador-server/src/prometheus_metrics.rs
@@ -83,7 +83,7 @@ impl PrometheusMetrics {
 
     // Creates and installs the prometheus exporter as global recorder
     // Only one recorder can be registered for the lifetime of the application
-    fn init_handle() -> PrometheusHandle {
+    pub fn init_handle() -> PrometheusHandle {
         let prom_builder = PrometheusBuilder::new();
         prom_builder
             .install_recorder()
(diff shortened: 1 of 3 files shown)


limitador-server/src/prometheus_metrics.rs
```diff
@@ -226,6 +282,32 @@ pub mod tests {
...
+        assert!(!metrics_output.contains("myLabel=\"user 1\""));
+        prometheus_metrics
+            .set_custom_labels(HashMap::from([(
+                "myLabel".to_string(),
+                Expression::parse("'user ' + descriptors[1].foobar").expect("Invalid expression!"),
+            )]))
+            .expect("Failed to set custom labels");
+        prometheus_metrics.incr_limited_calls(&namespace, None, &ctx);
+        let metrics_output = prometheus_metrics.gather_metrics();
+        assert!(metrics_output.contains("myLabel=\"user 1\""));
```
alexsnaps, line 308: Adds a label just fine, right?

alexsnaps:
@adam-cattermole this [last changeset](https://github.com/Kuadrant/limitador/pull/442/commits/75eaa3bf141c2c1b6c751b0fa8e2bc8677ec1883) might be of interest to you. Now I think there is more cleanup that probably could be done on the `PrometheusMetrics` itself and how it's built to avoid these issues in the future (some code still depends on that "global" recorder being wired in).

didierofrivia (approved):
Minor comment, feel free to merge without addressing.

PR #451, opened 2025-09-19
Bump rocksdb to 0.24

eguzki:
### What

Bump rocksdb to 0.24

### Why

`rocksdb` depends on librocksdb-sys` crate. Rocksdb 0.22 depends on librocksdb-sys 0.16.0. 

Recently I upgraded my Fedora's c++ compiler to 15. With c++ 15, librocksdb-sys 0.16.0 cannot be compiled, because `#include <cstdint>` is missing. Some C++ Standard Library headers have been modified to no longer include other headers that were previously used internally. As a result, C++ programs that rely on standard library components without explicitly including the appropriate headers will now fail to compile.

This has been fixed in  librocksdb-sys 0.17.x and now compiles.  

## Summary by CodeRabbit

* **Chores**
  * Updated an internal optional storage dependency to a newer stable release.

* **Bug Fixes**
  * Improved cache invalidation behavior for in-memory storage when invalidation closures are unavailable, making stale-counter invalidation more reliable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -570,29 +570,9 @@ checksum = "4372b9543397a4b86050cc5e7ee36953edf4bac9518e8a774c2da694977fb6e4"
 
 [[package]]
 name = "bindgen"
-version = "0.69.5"
+version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
-dependencies = [
- "bitflags 2.9.3",
- "cexpr",
- "clang-sys",
- "itertools 0.12.1",
- "lazy_static",
- "lazycell",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash 1.1.0",
- "shlex",
- "syn 2.0.106",
-]
-
-[[package]]
-name = "bindgen"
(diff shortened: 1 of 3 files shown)


eguzki:
There is a temporary issue for this PR. The required [librocksdb-sys 0.17.3](https://crates.io/crates/librocksdb-sys/0.17.3+10.4.2) version that has the fix to be compiled with c+`+ 15, requires rust 1.85.0.  [RHEL for the downstream only supports 1.84.1](https://docs.redhat.com/en/documentation/red_hat_developer_tools/1) as of today.

thomasmaas:
The Rust 1.85.0 blocker from September is resolved — limitador now builds with Rust 1.88 ([Dockerfile](https://github.com/Kuadrant/limitador/blob/main/Dockerfile#L1)). `librocksdb-sys 0.17.3` should compile fine.

This PR needs a rebase to resolve merge conflicts before it can be re-tested.

eguzki:
@coderabbitai review

eguzki:
@thomasmaas confirmed, it works with With c++ 15 (and c++ 16)

Needs approval

thomasmaas (approved):
Looks good — tight scope, valid motivation (C++15 compilation fix), and the `invalidate(c.as_ref())` change is correct. All CI green including multi-arch docker builds.

PR #453, opened 2025-09-22
GH action: build image as test

eguzki:
As part of the test suite run for each PR, add docker image build. Very useful for early caching of library dependency issues with supported rust version.


.github/workflows/rust.yml
```diff
@@ -79,12 +79,47 @@ jobs:
...
+      matrix:
+        include:
+          - dockerfile: Dockerfile
+            platform: linux/amd64
+            scope: build-amd
+            artifactsuffix: amd
+          - dockerfile: Dockerfile.aarch64
+            platform: linux/arm64
+            scope: build-arm
+            artifactsuffix: arm
```
KevFan, line 97: nit: these are not needed here since this test just builds the image and doesn't output to manifests or github cache

eguzki: Fixed. 

Ready for review  again.

PR #463, opened 2026-01-05
chore: updated CEL crate to v0.12

alexsnaps:


limitador-server/src/http_api/server.rs
```diff
@@ -439,7 +439,7 @@ mod tests {
         // Prepare values to check
         let mut values = HashMap::new();
         values.insert("req.method".into(), "GET".into());
-        values.insert("req.id".into(), "1".into());
+        values.insert("app.id".into(), "1".into());
```
alexsnaps: I'm confused how this test actually was _not_ failing before!?

alexsnaps: Ah! yes... @adam-cattermole the change where it now returns `NoSuchKey` instead of `Null`!
So that the lack of the var actually created a crap counter somewhere 🤦

Issue #466, opened 2026-01-22
In-memory cache for qualified counters need some lower limit


alexsnaps:
Today, if the system reports being maxed out, we happily create a "`0 bytes` cap'ed" cache for them, which means no qualified counters will ever be kept around and as such these limits end up effectively _not enforced_

alexsnaps:
Thanks for the PR, quickly looking at it, and while in agreement with @adam-cattermole there, I'm wondering if we shouldn't 🔥 this whole logic altogether. That math is flawed anyways (the size stuff)... This _could_ be an opportunity to rethink that heuristic altogether... 
Other than that, yes, if we can't come up with a good heuristic automatically, the server should probably _not even start_.

PR #467, opened 2026-01-29
Added minimum in-memory cache size

jrhyness:
**Description**
If the cache size is not specified via --cache and there is no available memory reported, it gets set to 0 and qualified counters cannot be cached.  This PR introduces a minimum cache size, as proposed in the linked issue. 

**Related Issue**
https://github.com/Kuadrant/limitador/issues/466

**Tests**
With no available memory reported:
```
{"timestamp":"2026-01-29T17:00:56.598806Z","level":"WARN","fields":{"message":"No cache size provided. Using 1000 entries (70% of 0MB available memory with a 1000 entry minimum)."},"target":"limitador_server"}
```

With available memory:
```
{"timestamp":"2026-01-29T17:03:42.177875Z","level":"WARN","fields":{"message":"No cache size provided. Using 144703856 entries (70% of 15771MB available memory with a 1000 entry minimum)."},"target":"limitador_server"}
```

### DIFF
--- a/limitador-server/src/main.rs
+++ b/limitador-server/src/main.rs
@@ -166,7 +166,7 @@ impl Limiter {
 
     fn in_memory_limiter(cfg: InMemoryStorageConfiguration) -> Self {
         let rate_limiter_builder =
-            RateLimiterBuilder::new(cfg.cache_size.or_else(guess_cache_size).unwrap());
+            RateLimiterBuilder::new(validate_cache_size(cfg.cache_size));
 
         Self::Blocking(rate_limiter_builder.build())
     }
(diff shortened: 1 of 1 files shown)


jrhyness:
> #466



> Thanks for the PR! For more context - this issue arose as on an overcommitted machine I had `free_size` report 0 and use a cache size of 0 (so all of my qualified counters were thrown away). My expectation was more so that if the `size` (70% of free mem) was below some threshold (unsure on the best value for this) we just exit out with an error.

I hit this as well.  In my case, on a Mac, `sys.available_memory()` reported 0 but `sys.free_memory()` was quite a bit more.   I'll revisit this.  Thanks again.

PR #500, opened 2026-07-21
Add two-phase release process with automated workflows

Boomatang:
# Summary

Replaces the manual, multi-step release process with an automated two-phase workflow driven by GitHub Actions and a `release.yaml` version manifest.

- **Phase 1 (Pre-release)**: A `workflow_dispatch` workflow creates a release branch, bumps versions in `release.yaml` and both `Cargo.toml` files, and opens a PR against the release branch. For minor releases it also opens a post-release PR to `main` with the next `-dev` versions.
- **Phase 2 (Release)**: A second `workflow_dispatch` workflow reads versions from `release.yaml`, runs smoke tests (fmt, clippy, full test suite, publish dry-run), creates annotated git tags, publishes the crate to crates.io, builds multi-arch container images, and creates the GitHub Release only after all artifacts succeed.
- **Version gate**: A new CI check on PRs targeting `release-*` branches validates that `release.yaml` contains non-sentinel, non-dev versions and that any declared dependency releases exist.
- **Reusable image build**: The existing `build-image.yaml` workflow now accepts `ref` and `tag` inputs via `workflow_call`, allowing the release workflow to trigger image builds for a specific ref and tag while preserving the existing push-on-main behavior.
- Adds `publish = false` to `limitador-server/Cargo.toml` to prevent accidental publication of the server binary crate.
- Introduces `release.yaml` at the repo root as the single source of truth for version numbers, using `0.0.0` as the sentinel value on `main`.
- Adds helper scripts (`parse-version.sh`, `validate-release-yaml.sh`) with documentation covering usage and error cases.
- Rewrites `RELEASE.md` to document the new process including minor releases, patch releases, rollback procedures, and file inventory.

## Examples
Pre-release Workflow: https://github.com/Boomatang/limitador/actions/runs/28022410045
PR to release branch: https://github.com/Boomatang/limitador/pull/5
PR to main (dev version bump): https://github.com/Boomatang/limitador/pull/6
Release Workflow: https://github.com/Boomatang/limitador/actions/runs/28028029786
GitHub release: https://github.com/Boomatang/limitador/releases

## Summary by CodeRabbit

- **New Features**
  - Added automated pre-release and release workflows with semver parsing, version/branch validation, tagging, crates.io dry-run, image builds, and GitHub Release creation.
  - Introduced a version-gate workflow for pull requests to keep `release.yaml` and Cargo versions in sync.
  - Added reusable image-build support with optional tagging.
  - Added release helper scripts and a composite action to install `yq`.
- **Documentation**
  - Updated `RELEASE.md` with the new two-phase process, conventions, and staged rollback guidance.
  - Added documentation for the release helper scripts.
- **Configuration**
  - Disabled publication of the server crate.
  - Added initial `release.yaml` metadata configuration.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -2,12 +2,20 @@
 name: Build Image
 
 on:
+  workflow_call:
+    inputs:
+      ref:
+        description: "Git ref to check out"
+        required: true
+        type: string
+      tag:
+        description: "Image tag to apply"
+        required: true
+        type: string
   workflow_dispatch:
   push:
     branches:
       - main
-    tags:
-      - "*"
 
 env:
   IMG_REGISTRY_HOST: quay.io
@@ -37,6 +45,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.ref || github.ref }}
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
       - name: Docker meta
@@ -99,12 +109,14 @@ jobs:
           images: |
             ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador
           tags: |
+            # When called from release workflow, use the provided tag
+            type=raw,value=${{ inputs.tag }},enable=${{ inputs.tag != '' }}
             # SHA tag for main branch
-            type=raw,value=${{ github.sha }},enable=${{ github.ref_name == env.MAIN_BRANCH_NAME }}
+            type=raw,value=${{ github.sha }},enable=${{ inputs.tag == '' && github.ref_name == env.MAIN_BRANCH_NAME }}
             # set latest tag for main branch
-            type=raw,value=latest,enable=${{ github.ref_name == env.MAIN_BRANCH_NAME }}
+            type=raw,value=latest,enable=${{ inputs.tag == '' && github.ref_name == env.MAIN_BRANCH_NAME }}
             # set ref name tag for non-main branches
-            type=raw,value=${{ github.ref_name }},enable=${{ github.ref_name != env.MAIN_BRANCH_NAME }}
+            type=raw,value=${{ github.ref_name }},enable=${{ inputs.tag == '' && github.ref_name != env.MAIN_BRANCH_NAME }}
       - name: Login to container registry
         uses: docker/login-action@v2
         with:
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -55,6 +55,9 @@ jobs:
         with:
           images: |
             ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador
+      - name: Resolve checked-out SHA
+        id: sha
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
       - name: Login to container registry
         uses: docker/login-action@v2
         with:
@@ -68,7 +71,7 @@ jobs:
           context: .
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
-            GITHUB_SHA=${{ github.sha }}
+            GITHUB_SHA=${{ steps.sha.outputs.sha }}
           cache-from: type=gha,scope=${{ matrix.scope }}
           cache-to: type=gha,mode=max,scope=${{ matrix.scope}}
           outputs: type=image,name=${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/limitador,push-by-digest=true,name-canonical=true,push=true
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,281 @@
+---
+name: Pre-release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Server release version (semver, e.g. 2.5.0)"
+        required: true
+        type: string
+      crate-version:
+        description: "Crate release version (semver, e.g. 0.13.0)"
+        required: true
+        type: string
+      source-branch:
+        description: "Branch to base the pre-release on (default: main for new releases, release branch for patches)"
+        required: false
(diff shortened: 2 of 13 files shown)


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,281 @@
...
+        run: |
+          yq --inplace ".limitador.version = \"${VERSION}\"" release.yaml
+          yq --inplace ".limitador.\"crate-version\" = \"${CRATE_VERSION}\"" release.yaml
+
+      - name: Update Cargo.toml versions
+        env:
+          VERSION: ${{ needs.setup.outputs.version }}
+          CRATE_VERSION: ${{ needs.setup.outputs.crate-version }}
+        run: |
+          sed --in-place '/^\[package\]/,/^\[/{s/^version = ".*"/version = "'"${CRATE_VERSION}"'"/}' limitador/Cargo.toml
```
eguzki: Consider replacing sed range-address version bumps with cargo set-version (from cargo-edit)

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,281 @@
...
+            exit 1
+          fi
+          git switch --create "${POST_RELEASE_BRANCH}"
+
+      - name: Update Cargo.toml versions
+        env:
+          NEXT_VERSION: ${{ steps.next.outputs.version }}
+          NEXT_CRATE_VERSION: ${{ steps.next.outputs.crate-version }}
+        run: |
+          sed --in-place '/^\[package\]/,/^\[/{s/^version = ".*"/version = "'"${NEXT_CRATE_VERSION}"'"/}' limitador/Cargo.toml
```
eguzki: Consider replacing sed range-address version bumps with cargo set-version (from cargo-edit)

eguzki (commented):
The overall approach looks good to me. The release process looks more robust.

Before approval, let me ask. Would it make sense to implement this 2-phase release process without release.yaml file? This eliminates the drift risk where tags and github releases (from release.yaml) could disagree with what cargo publish actually ships(from Cargo.toml). Actually, I am suggesting moving the version source of truth from release.yaml to both Cargo.toml. The `read-version` job would parse `limitador-server/Cargo.toml` / `limitador/Cargo.toml` directly instead of yq parsing release.yaml. . Drop the 0.0.0 sentinel scheme in favor of the existing -dev suffix convention already used in Cargo.toml on main. Move version-gate.yaml's sentinel/dev-suffix checks to inspect Cargo.toml instead of release.yaml. Shrink release.yaml down to just the `dependencies: {}` stanza. BTW, what is the purpose of `dependencies` in this repo?

.github/workflows/pre-release.yaml
Boomatang: I was unaware of cargo-edit. I will of course change that.

Boomatang:
> Would it make sense to implement this 2-phase release process without release.yaml file?

The release.yaml will be used by external tooling also. All kuadrant repos will have the same file, so automating build the tree of version from the kuadrant-operator becomes much simpler. As that file will be in the root of all projects, on all branches. So to find the version of this rust project, has the same logic as a go, or typescript project.  

> BTW, what is the purpose of dependencies in this repo?

While not really needed here, it is used more for building the up the tree of dependencies, and the RFC calls out that field being there. It could be removed.

> the drift risk where tags and github releases could disagree with what cargo publish actually ships

This is a concern, but with not having any workflows triggered outside release process related to the process I think cuts down on the risk. This drift risk is an issue across all repos, and different languages.

> Drop the 0.0.0 sentinel scheme in favor of the existing -dev suffix convention already used in Cargo.toml on main.

It is not a convention used in the go based projects. I think the 0.0.0 also indicates it is `main` and not some patch that is in progress. For example what branch would you expect 1.10.1-dev to point to. For me I would would think it is release-1.10. But I know that is a small argument, and it would be easy to build around.

eguzki:
Thanks for the context on external tooling, that changes my read. I don't think we need to drop `release.yaml`, just flip which file is authoritative. 

Consider the following proposal:

**Source of truth: `Cargo.toml`.** `release.yaml` stays at the repo root with the same shape, so `kuadrant-operator` (and anything else walking the tree) keeps reading one uniform file regardless of language. It just becomes a generated mirror instead of the thing people hand-edit. This is what actually eliminates the drift risk I flagged (tags/GitHub releases disagreeing with what `cargo publish` ships) `release.yaml` can never say something `Cargo.toml` doesn't.

**How the mirror gets updated:** The pre-release and release pipelines are the only places versions change, so they write `Cargo.toml` first (as `pre-release.yaml` already does) and then generate `release.yaml` from it in that same controlled step.

**Version gate, independent of the pipelines:** a check that `release.yaml` and `Cargo.toml` agree. Running on all branches (not only in `main` and `release-*`). Someone bumping a `Cargo.toml` version directly wouldn't be caught until a release-branch PR forced the issue. 

**Allow `main`'s version drift (`0.0.0` in release.yaml vs `-dev` in cargo.toml)**

- If `release.yaml` version is `0.0.0`, the only valid pairing is a `Cargo.toml` version ending in `-dev`. That's the one allowed exception.
- Otherwise, `release.yaml` must equal `Cargo.toml` exactly. No drift permitted.

If this lands, I'd suggest it as the template for the other Rust-based Kuadrant repos too: native manifest as source of truth, `release.yaml` as a generated, gate-verified mirror for cross-repo tooling.

.github/workflows/build-image.yaml
eguzki: We might have an issue here. 

When release.yaml calls build-image.yaml via workflow_call, the reusable workflow inherits the same run context, same github.sha, same github.ref. So, if the release workflow is run from `main`, the image gets mislabeled with main's SHA instead of the release branch's commit.The GitHub Actions UI defaults "Use workflow from" to the repository's default branch (main) every time you open the dispatch dialog. The doc does not explicitly state to pick the release branch. (seems obvious, but not stated). 

Would it be safer to do the following?

```yaml
- name: Resolve checked-out SHA
  id: sha
  run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"

- name: Build and push by digest
  ...
  with:
    build-args: |
      GITHUB_SHA=${{ steps.sha.outputs.sha }}
```

eguzki (approved):

PR #501, opened 2026-07-22
fix: replace .unwrap() calls with descriptive .expect() messages

### LINKED ISSUE #280: Clean up `Result.unwrap` usages
Adding the lint to our crate and server code should help finding and going about them

```diff
diff --git a/limitador-server/src/main.rs b/limitador-server/src/main.rs

-#![deny(clippy::all, clippy::cargo)]
+#![deny(clippy::all, clippy::cargo, clippy::unwrap_used)]
 
diff --git a/limitador/src/lib.rs b/limitador/src/lib.rs
 
-#![deny(clippy::all, clippy::cargo)]
+#![deny(clippy::all, clippy::cargo, clippy::unwrap_used)]
```

Not that all necessary have to go, but probably cleaning some of these up and possibly swapping them for `.expect()` would be desirable.

Taimoo-r:
Closes #280

## What this does

Enables the `clippy::unwrap_used` lint at crate level in both `limitador` and `limitador-server`:

```rust
#![deny(clippy::all, clippy::cargo, clippy::unwrap_used)]
```

Every `.unwrap()` call across both crates is replaced with `.expect("...")` carrying a message that explains the invariant or assumption behind it. Categories addressed:

- **RwLock poisoning** (`read/write().unwrap()`) - replaced with `.expect("lock poisoned")`
- **SystemTime before Unix epoch** - replaced with `.expect("system time before Unix epoch")`
- **Serialization/deserialization** (serde_json, postcard) - replaced with `.expect("serialization failed")` / `.expect("key deserialization failed")`
- **Post-error-check unwraps** (called after an early return on `Err`) - replaced with descriptive invariant messages
- **CLI required args** (clap `get_one` on required args) - replaced with `.expect("... arg is required")`
- **Test modules** - `#[allow(clippy::unwrap_used)]` added to each `#[cfg(test)] mod tests` block so test code is unaffected

## Testing

- `cargo clippy --no-default-features -p limitador` passes with zero errors
- The `disk_storage` and `redis_storage` features require libclang and Perl respectively and cannot be compiled on the development machine (Windows); those modules were fixed by reading source and applying the same patterns, and CI on Linux will validate them

## Summary by CodeRabbit

* **Reliability**
  * Improved diagnostic messages for configuration, storage, networking, locking and server-startup failures.
  * Added clearer reporting for invalid state, encoding, time and address-handling errors.

* **Quality**
  * Strengthened static checks to detect unsafe unchecked operations during development.
  * Existing test coverage remains supported under the stricter checks.

* **Compatibility**
  * No public APIs, rate-limiting behaviour or data formats were changed.

### DIFF
--- a/limitador-server/src/envoy_rls/kuadrant_service.rs
+++ b/limitador-server/src/envoy_rls/kuadrant_service.rs
@@ -78,7 +78,7 @@ impl RateLimitService for KuadrantService {
             return Err(Status::unavailable("Service unavailable"));
         }
 
-        let rate_limited_resp = rate_limited_resp.unwrap();
+        let rate_limited_resp = rate_limited_resp.expect("rate_limited_resp must be Ok after error check");
         let resp_code = if rate_limited_resp.limited {
             self.metrics.incr_limited_calls(
                 &namespace,
@@ -186,6 +186,7 @@ impl RateLimitService for KuadrantService {
 }
 
 #[cfg(test)]
+#[allow(clippy::unwrap_used)]
 mod tests {
     mod check_rate_limit {
         use tonic::IntoRequest;
--- a/limitador-server/src/envoy_rls/server.rs
+++ b/limitador-server/src/envoy_rls/server.rs
@@ -171,7 +171,7 @@ impl RateLimitService for MyRateLimiter {
             return Err(Status::unavailable("Service unavailable"));
         }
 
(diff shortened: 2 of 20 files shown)


Taimoo-r:
Hi @jasonmadigan, could you please look into this. Thanks

PR #515, opened 2026-09-06
506 implement token rate limit reservations reservecommit rpcs

### LINKED ISSUE #506: Implement token rate limit reservations (Reserve/Commit RPCs)
## Summary
Add a reservation layer to Limitador: hold estimated token capacity on arrival (`Reserve`), release/finalize it with actual usage when the response is known (`Commit`). Closes the race window where concurrent in-flight requests can pass a `hits_addend=0` check before any of them report usage.

Design: [RFC 0021](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md), reference-level section "Limitador".

## Acceptance Criteria
- [ ] `ReservationEntry` type and per-counter `CounterReservationRegistry`, keyed by the same `(limit, set_variables)` identity `Counter` already uses
- [ ] Reservation persistence tiered by backend:
  - [ ] `InMemoryStorage`: dedicated Moka cache using Moka's per-entry `Expiry` trait
  - [ ] `RedisStorage`/`AsyncRedisStorage`/`CachedRedisStorage`: hash-field-backed (`amount` + `expires_at`), shared across replicas, with key-level `EXPIRE`
  - [ ] `RocksDbStorage` and the gossip-based backend: local-memory-only (documented limitation, not a regression)
- [ ] `RateLimiter::reserve` / `RateLimiter::commit_reservation` (+ async twins), composed from existing `counters_that_apply` / `update_counter` / admission-check logic
- [ ] Admission check accounts for outstanding reservations: `counter.value + outstanding(counter) + requested_amount <= limit`
- [ ] TTL/expiry: `expires_at(counter) = min(now + reservation.ttl, current_window_expiry(counter), now + max_reservation_ttl)`, computed independently per counter
- [ ] `Reserve` / `Commit` proto messages and RPCs added to the existing Kuadrant gRPC service, alongside (not replacing) `CheckRateLimit`/`Report`
- [ ] `Commit` re-resolves counters from `descriptors`, always applies `actual_amount` unconditionally, and additionally looks up/removes the `reservation_id` entry if present and live (degrades gracefully to `Report` behavior otherwise)
- [ ] `--disable-reservations` server flag (default `false`); when set, `Reserve`/`Commit` are unregistered or return `UNIMPLEMENTED`
- [ ] `--max-reservation-fraction` server flag (default `0.5`): clamps `requested_amount(counter) = min(caller_amount, limit(counter) * fraction)`
- [ ] `--max-reservation-ttl` server flag (default `60s`): hard ceiling on `reservation.ttl`
- [ ] Concurrency/race tests directly exercising the scenario in RFC 0021's Motivation (N concurrent requests, one limit)

## Notes for reviewers
- Standard Envoy RLS v3 `ShouldRateLimit` is untouched; this only exists on the custom Kuadrant service.
- PR description should include: how to review (suggested order: registry → admission check → RPCs → flags → tests) and manual verification steps (e.g. local Limitador + grpcurl `Reserve`/`Commit` calls showing admission and release).

eguzki:
Fixes #506

### Verification steps

Run the `sandbox` dev environment:

```bash
cd limitador-server/sandbox
docker rmi limitador-testing:latest
make build
make deploy-in-memory
```

In a separate shell, install [grpcurl](https://github.com/fullstorydev/grpcurl) (needs [Go SDK 1.18+](https://golang.org/doc/install)):

```bash
make grpcurl
```

Inspect the `RateLimitService` GRPC service — it should now list `Reserve` and `Commit` alongside `CheckRateLimit`/`Report`:

```bash
bin/grpcurl -plaintext 127.0.0.1:18081 describe kuadrant.service.ratelimit.v1.RateLimitService
```
```
kuadrant.service.ratelimit.v1.RateLimitService is a service:
service RateLimitService {
  rpc CheckRateLimit ( .envoy.service.ratelimit.v3.RateLimitRequest ) returns ( .envoy.service.ratelimit.v3.RateLimitResponse );
  rpc Report ( .envoy.service.ratelimit.v3.RateLimitRequest ) returns ( .envoy.service.ratelimit.v3.RateLimitResponse );
  rpc Reserve ( .kuadrant.service.ratelimit.v1.ReserveRequest ) returns ( .kuadrant.service.ratelimit.v1.ReserveResponse );
  rpc Commit ( .kuadrant.service.ratelimit.v1.CommitRequest ) returns ( .kuadrant.service.ratelimit.v1.CommitResponse );
}
```

The sandbox's `limits.yaml` has a `GET`/`req.path: /` limit of max 10 req in a 60s window. Reserve 1 unit of estimated capacity against it:

```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 kuadrant.service.ratelimit.v1.RateLimitService.Reserve <<EOM
{
    "domain": "test_namespace",
    "amount": 1,
    "ttl": "60s",
    "descriptors": [
        {"entries": [
            {"key": "req.method", "value": "GET"},
            {"key": "req.path", "value": "/"}
        ]}
    ]
}
EOM
```

Expected response — `OK`, with a `reservationId` to keep for later:

```json
{
  "code": "OK",
  "reservationId": "<uuid>"
}
```

Note the counter itself is untouched by the reservation alone — `remaining` still shows 10, since the hold lives separately from the counter's own value until `Commit` applies it:

```bash
curl -s http://127.0.0.1:18080/counters/test_namespace | yq e -P
```

Reserve 1 more unit, 9 more times. Even though the counter's own value is still 0, the outstanding holds from these (plus the one above) fill the limit of 10 - the 10th call in this loop is the last one that should still be `OK`:

```bash
for i in $(seq 1 9); do
  bin/grpcurl -plaintext -d @ 127.0.0.1:18081 kuadrant.service.ratelimit.v1.RateLimitService.Reserve <<EOM
{
    "domain": "test_namespace",
    "amount": 1,
    "ttl": "60s",
    "descriptors": [{"entries": [{"key": "req.method", "value": "GET"}, {"key": "req.path", "value": "/"}]}]
}
EOM
done
```

One more `Reserve` call now returns `OVER_LIMIT`, since outstanding reservations already total 10:

```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 kuadrant.service.ratelimit.v1.RateLimitService.Reserve <<EOM
{
    "domain": "test_namespace",
    "amount": 1,
    "ttl": "60s",
    "descriptors": [
        {"entries": [
            {"key": "req.method", "value": "GET"},
            {"key": "req.path", "value": "/"}
        ]}
    ]
}
EOM
```
```json
{
  "code": "OVER_LIMIT"
}
```

Resolve the first reservation with `Commit`, reporting real usage:

```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 kuadrant.service.ratelimit.v1.RateLimitService.Commit <<EOM
{
    "domain": "test_namespace",
    "reservationId": "<paste the reservation_id here>",
    "actualAmount": 1,
    "descriptors": [
        {"entries": [
            {"key": "req.method", "value": "GET"},
            {"key": "req.path", "value": "/"}
        ]}
    ]
}
EOM
```

Expected response, and the reservation's hold is now released:

```json
{
  "reservationReleased": true
}
```

Committing the *same* `reservationId` again (already released) degrades gracefully: `actualAmount` is still applied to the counter unconditionally, but nothing is released this time. The expected response is an **empty object**, not `{"reservationReleased": false}`:

```bash
bin/grpcurl -plaintext -d @ 127.0.0.1:18081 kuadrant.service.ratelimit.v1.RateLimitService.Commit <<EOM
{
    "domain": "test_namespace",
    "reservationId": "<the same reservation_id again>",
    "actualAmount": 1,
    "descriptors": [
        {"entries": [
            {"key": "req.method", "value": "GET"},
            {"key": "req.path", "value": "/"}
        ]}
    ]
}
EOM
```
```json
{}
```

The counter now reflects the committed usage:

```bash
curl -s http://127.0.0.1:18080/counters/test_namespace | yq e -P
```
```yaml
- limit:
    namespace: test_namespace
    max_value: 10
    seconds: 60
    conditions:
      - descriptors[0]['req.method'] == 'GET'
      - descriptors[0]['req.path'] != '/json'
    variables: []
  set_variables: {}
  remaining: 8
  expires_in_seconds: <=60
```

### Clean up
```
make clean
```

## Summary by CodeRabbit

* **New Features**
  * Added rate-limit reservations, allowing estimated capacity to be held and later committed using actual usage.
  * Added Reserve and Commit gRPC operations with reservation identifiers, expiry handling, and configurable limits.
  * Added command-line options to disable reservations and configure reservation fraction and TTL.
  * Added reservation support across in-memory, disk, Redis, and cached Redis storage.

* **Documentation**
  * Added sandbox examples for reserving and committing capacity.
  * Clarified that testing requires Redis 7.0 or newer.

* **Chores**
  * Updated sandbox Redis images to Redis 7.

### DIFF
--- a/limitador-server/proto/kuadrantrls.proto
+++ b/limitador-server/proto/kuadrantrls.proto
@@ -34,8 +34,10 @@ message ReserveRequest {
 
 message ReserveResponse {
   envoy.service.ratelimit.v3.RateLimitResponse.Code code = 1;
-  // Only set when code is OK. Must be passed back to Commit to resolve this reservation.
-  string reservation_id = 2;
+  // Set when code is OK and something was actually held. Must be passed back to Commit to
+  // resolve this reservation; absent means there's nothing to resolve (e.g. the amount held
+  // ended up being 0).
+  optional string reservation_id = 2;
   // Only set when code is OK. The amount actually held - can be less than the requested
   // amount if the server clamped it down; purely informational, Commit's actual_amount is
   // independent of it.
@@ -45,7 +47,9 @@ message ReserveResponse {
 message CommitRequest {
   string domain = 1;
   repeated envoy.extensions.common.ratelimit.v3.RateLimitDescriptor descriptors = 2;
-  string reservation_id = 3;
+  // Absent degrades gracefully to Report-equivalent behavior, same as an unrecognized or
+  // already-expired id.
+  optional string reservation_id = 3;
   uint64 actual_amount = 4;
 }
 
--- a/limitador-server/proto/kuadrantrls.proto
+++ b/limitador-server/proto/kuadrantrls.proto
@@ -3,11 +3,48 @@ syntax = "proto3";
 package kuadrant.service.ratelimit.v1;
 
 import "envoy/service/ratelimit/v3/rls.proto";
+import "envoy/extensions/common/ratelimit/v3/ratelimit.proto";
+import "google/protobuf/duration.proto";
 
 service RateLimitService {
   // Determine whether a request should be allowed or denied based on rate limiting rules.
-  rpc CheckRateLimit(envoy.service.ratelimit.v3.RateLimitRequest) returns (envoy.service.ratelimit.v3.RateLimitResponse); 
+  rpc CheckRateLimit(envoy.service.ratelimit.v3.RateLimitRequest) returns (envoy.service.ratelimit.v3.RateLimitResponse);
 
   // The report method modifies specific counters associated with the descriptors, incrementing each one by the amount specified in $hits_addend$
-  rpc Report(envoy.service.ratelimit.v3.RateLimitRequest) returns (envoy.service.ratelimit.v3.RateLimitResponse); 
+  rpc Report(envoy.service.ratelimit.v3.RateLimitRequest) returns (envoy.service.ratelimit.v3.RateLimitResponse);
+
+  // Reserve holds estimated capacity (e.g. estimated tokens) against the matching counters,
+  // without exceeding their limits once other outstanding reservations are accounted for.
+  // The hold lasts until Commit resolves it or it expires, whichever comes first.
+  rpc Reserve(ReserveRequest) returns (ReserveResponse);
+
+  // Commit resolves a reservation with the caller's real usage: it always applies
+  // actual_amount to the matching counters, and releases the reservation_id's hold if it is
+  // still found and live. A late or already-expired reservation_id degrades gracefully to
+  // Report-equivalent behavior.
+  rpc Commit(CommitRequest) returns (CommitResponse);
+}
+
+message ReserveRequest {
+  string domain = 1;
+  repeated envoy.extensions.common.ratelimit.v3.RateLimitDescriptor descriptors = 2;
+  uint64 amount = 3;
+  google.protobuf.Duration ttl = 4;
+}
+
+message ReserveResponse {
+  envoy.service.ratelimit.v3.RateLimitResponse.Code code = 1;
+  // Only set when code is OK. Must be passed back to Commit to resolve this reservation.
+  string reservation_id = 2;
+}
+
+message CommitRequest {
+  string domain = 1;
+  repeated envoy.extensions.common.ratelimit.v3.RateLimitDescriptor descriptors = 2;
+  string reservation_id = 3;
+  uint64 actual_amount = 4;
+}
+
+message CommitResponse {
+  bool reservation_released = 1;
 }
--- a/limitador-server/proto/kuadrantrls.proto
+++ b/limitador-server/proto/kuadrantrls.proto
@@ -36,6 +36,10 @@ message ReserveResponse {
   envoy.service.ratelimit.v3.RateLimitResponse.Code code = 1;
   // Only set when code is OK. Must be passed back to Commit to resolve this reservation.
   string reservation_id = 2;
+  // Only set when code is OK. The amount actually held - can be less than the requested
+  // amount if the server clamped it down; purely informational, Commit's actual_amount is
+  // independent of it.
+  uint64 reserved_amount = 3;
 }
 
 message CommitRequest {
--- a/limitador-server/src/envoy_rls/kuadrant_service.rs
+++ b/limitador-server/src/envoy_rls/kuadrant_service.rs
@@ -10,26 +10,42 @@ use super::server::custom::service::ratelimit::v1::{
 };
 use super::server::envoy::service::ratelimit::v3::rate_limit_response::Code;
 use super::server::envoy::service::ratelimit::v3::{RateLimitRequest, RateLimitResponse};
+use super::server::ReservationConfig;
 use crate::prometheus_metrics::PrometheusMetrics;
 use crate::Limiter;
 use limitador::limit::Context;
 use limitador::reservation::ReservationId;
 
-// RFC 0021 defaults. Not yet wired to the `--max-reservation-ttl`/`--max-reservation-fraction`
-// server flags (separate, not-yet-implemented follow-up) - a caller-requested ttl is clamped
-// to this ceiling, and it's also used when the caller doesn't set `ttl` at all. There is no
-// equivalent clamp yet for `amount` (the fraction-of-limit clamp needs each matching counter's
-// own max_value, which isn't known until inside `RateLimiter::reserve`).
-const MAX_RESERVATION_TTL: Duration = Duration::from_secs(60);
+// Used only when the caller doesn't set `ttl` at all. Deliberately not "the real default" -
+// the library's own `RateLimiterBuilder::reservation_limits` ceiling (set once, at
+// construction time) always clamps this down to whatever's actually configured, so this only
+// needs to be "large enough to never be the binding constraint," not accurate.
+const UNSET_RESERVATION_TTL: Duration = Duration::from_secs(u32::MAX as u64);
 
 pub struct KuadrantService {
     limiter: Arc<Limiter>,
     metrics: Arc<PrometheusMetrics>,
+    reservation_config: ReservationConfig,
 }
 
 impl KuadrantService {
+    // Only used by tests: production code (`server.rs`) always goes through
+    // `new_with_reservation_config` so it can pass along the real, CLI-configured settings.
+    #[cfg(test)]
     pub fn new(limiter: Arc<Limiter>, metrics: Arc<PrometheusMetrics>) -> Self {
-        Self { limiter, metrics }
+        Self::new_with_reservation_config(limiter, metrics, ReservationConfig::default())
+    }
+
+    pub fn new_with_reservation_config(
+        limiter: Arc<Limiter>,
+        metrics: Arc<PrometheusMetrics>,
+        reservation_config: ReservationConfig,
+    ) -> Self {
+        Self {
+            limiter,
+            metrics,
+            reservation_config,
+        }
     }
 }
 
@@ -203,6 +219,10 @@ impl RateLimitService for KuadrantService {
     ) -> Result<Response<ReserveResponse>, Status> {
         debug!("Reserve request received: {:?}", request);
 
+        if self.reservation_config.disable_reservations {
+            return Err(Status::unimplemented("reservations are disabled"));
+        }
+
         let mut values: Vec<HashMap<String, String>> = Vec::default();
         let (_metadata, _ext, req) = request.into_parts();
         let namespace = req.domain;
@@ -227,11 +247,16 @@ impl RateLimitService for KuadrantService {
         let mut ctx = Context::default();
         ctx.list_binding("descriptors".to_string(), values);
 
+        // Both the fraction-of-limit clamp on `amount` and the ceiling on `ttl` happen inside
+        // `RateLimiter::reserve` itself (`RateLimiterBuilder::reservation_limits`, set once
+        // when the limiter is constructed) - the fraction clamp needs each matching counter's
+        // own max_value, already resolved there, so there's no point re-resolving limits here
+        // just to clamp again; `ttl` is simply passed through unclamped (or, if unset, as a
+        // large sentinel the library's own ceiling will cut down to size).
         let ttl = req
             .ttl
             .and_then(|d| Duration::try_from(d).ok())
-            .unwrap_or(MAX_RESERVATION_TTL)
-            .min(MAX_RESERVATION_TTL);
+            .unwrap_or(UNSET_RESERVATION_TTL);
 
         let reserve_resp = match &*self.limiter {
             Limiter::Blocking(limiter) => limiter.reserve(&namespace, &ctx, req.amount, ttl, false),
@@ -276,6 +301,10 @@ impl RateLimitService for KuadrantService {
         request: Request<CommitRequest>,
     ) -> Result<Response<CommitResponse>, Status> {
         debug!("Commit request received: {:?}", request);
+
+        if self.reservation_config.disable_reservations {
+            return Err(Status::unimplemented("reservations are disabled"));
+        }
 
         let mut values: Vec<HashMap<String, String>> = Vec::default();
         let (_metadata, _ext, req) = request.into_parts();
--- a/limitador-server/src/envoy_rls/kuadrant_service.rs
+++ b/limitador-server/src/envoy_rls/kuadrant_service.rs
@@ -225,6 +225,7 @@ impl RateLimitService for KuadrantService {
             return Ok(Response::new(ReserveResponse {
                 code: Code::Unknown.into(),
                 reservation_id: String::new(),
+                reserved_amount: 0,
             }));
         }
 
@@ -277,6 +278,7 @@ impl RateLimitService for KuadrantService {
         Ok(Response::new(ReserveResponse {
             code: code.into(),
             reservation_id,
+            reserved_amount: reserve_resp.amount,
         }))
     }
 
@@ -952,6 +954,7 @@ mod tests {
                 .into_inner();
             assert_eq!(response.code, i32::from(Code::Ok));
             assert!(!response.reservation_id.is_empty());
+            assert_eq!(response.reserved_amount, 6);
 
             // Still held: 0 + outstanding(6) + 6 = 12 > 10
             let blocked = service
@@ -961,6 +964,7 @@ mod tests {
                 .into_inner();
             assert_eq!(blocked.code, i32::from(Code::OverLimit));
             assert!(blocked.reservation_id.is_empty());
+            assert_eq!(blocked.reserved_amount, 0);
 
             let commit_req = CommitRequest {
                 domain: namespace.to_string(),
--- a/limitador-server/src/envoy_rls/kuadrant_service.rs
+++ b/limitador-server/src/envoy_rls/kuadrant_service.rs
@@ -1,14 +1,26 @@
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::Duration;
 
 use tonic::{Request, Response, Status};
 
 use super::server::custom::service::ratelimit::v1::rate_limit_service_server::RateLimitService;
+use super::server::custom::service::ratelimit::v1::{
+    CommitRequest, CommitResponse, ReserveRequest, ReserveResponse,
+};
 use super::server::envoy::service::ratelimit::v3::rate_limit_response::Code;
 use super::server::envoy::service::ratelimit::v3::{RateLimitRequest, RateLimitResponse};
 use crate::prometheus_metrics::PrometheusMetrics;
 use crate::Limiter;
 use limitador::limit::Context;
+use limitador::reservation::ReservationId;
+
+// RFC 0021 defaults. Not yet wired to the `--max-reservation-ttl`/`--max-reservation-fraction`
+// server flags (separate, not-yet-implemented follow-up) - a caller-requested ttl is clamped
+// to this ceiling, and it's also used when the caller doesn't set `ttl` at all. There is no
+// equivalent clamp yet for `amount` (the fraction-of-limit clamp needs each matching counter's
+// own max_value, which isn't known until inside `RateLimiter::reserve`).
+const MAX_RESERVATION_TTL: Duration = Duration::from_secs(60);
 
 pub struct KuadrantService {
     limiter: Arc<Limiter>,
@@ -182,6 +194,138 @@ impl RateLimitService for KuadrantService {
         };
 
         Ok(Response::new(reply))
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn reserve(
+        &self,
+        request: Request<ReserveRequest>,
+    ) -> Result<Response<ReserveResponse>, Status> {
+        debug!("Reserve request received: {:?}", request);
+
+        let mut values: Vec<HashMap<String, String>> = Vec::default();
+        let (_metadata, _ext, req) = request.into_parts();
+        let namespace = req.domain;
+
+        if namespace.is_empty() {
+            return Ok(Response::new(ReserveResponse {
+                code: Code::Unknown.into(),
+                reservation_id: String::new(),
+            }));
+        }
+
+        let namespace = namespace.into();
+
+        for descriptor in &req.descriptors {
+            let mut map = HashMap::default();
+            for entry in &descriptor.entries {
+                map.insert(entry.key.clone(), entry.value.clone());
+            }
+            values.push(map);
+        }
+
+        let mut ctx = Context::default();
+        ctx.list_binding("descriptors".to_string(), values);
+
+        let ttl = req
+            .ttl
+            .and_then(|d| Duration::try_from(d).ok())
+            .unwrap_or(MAX_RESERVATION_TTL)
+            .min(MAX_RESERVATION_TTL);
+
+        let reserve_resp = match &*self.limiter {
+            Limiter::Blocking(limiter) => limiter.reserve(&namespace, &ctx, req.amount, ttl, false),
+            Limiter::Async(limiter) => {
+                limiter
+                    .reserve(&namespace, &ctx, req.amount, ttl, false)
+                    .await
+            }
+        };
+
+        if let Err(e) = reserve_resp {
+            // See the comment on the same pattern in `check_rate_limit` above: an
+            // "unavailable" error lets `failure_mode_deny` decide the outcome, rather than
+            // silently letting the request through.
+            error!("Error: {:?}", e);
+            return Err(Status::unavailable("Service unavailable"));
+        }
+
+        let reserve_resp = reserve_resp.unwrap();
+        let (code, reservation_id) = if reserve_resp.limited {
+            self.metrics
+                .incr_limited_calls(&namespace, reserve_resp.limit_name.as_deref(), &ctx);
+            (Code::OverLimit, String::new())
+        } else {
+            self.metrics.incr_authorized_calls(&namespace, &ctx);
+            let reservation_id = reserve_resp
+                .reservation_id
+                .map(|id| id.to_string())
+                .unwrap_or_default();
+            (Code::Ok, reservation_id)
+        };
+
+        Ok(Response::new(ReserveResponse {
+            code: code.into(),
+            reservation_id,
+        }))
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn commit(
+        &self,
+        request: Request<CommitRequest>,
+    ) -> Result<Response<CommitResponse>, Status> {
+        debug!("Commit request received: {:?}", request);
+
+        let mut values: Vec<HashMap<String, String>> = Vec::default();
+        let (_metadata, _ext, req) = request.into_parts();
+        let namespace = req.domain;
+
+        if namespace.is_empty() {
+            return Ok(Response::new(CommitResponse {
+                reservation_released: false,
+            }));
+        }
+
+        let namespace = namespace.into();
+
+        for descriptor in &req.descriptors {
+            let mut map = HashMap::default();
+            for entry in &descriptor.entries {
+                map.insert(entry.key.clone(), entry.value.clone());
+            }
+            values.push(map);
+        }
+
+        let mut ctx = Context::default();
+        ctx.list_binding("descriptors".to_string(), values);
+
+        let reservation_id = ReservationId::from(req.reservation_id);
+
+        let commit_resp = match &*self.limiter {
+            Limiter::Blocking(limiter) => {
+                limiter.commit_reservation(&namespace, &ctx, &reservation_id, req.actual_amount)
+            }
+            Limiter::Async(limiter) => {
+                limiter
+                    .commit_reservation(&namespace, &ctx, &reservation_id, req.actual_amount)
+                    .await
+            }
+        };
+
+        if let Err(e) = commit_resp {
+            error!("Error: {:?}", e);
+            return Err(Status::unavailable("Service unavailable"));
+        }
+
+        let commit_resp = commit_resp.unwrap();
+        self.metrics.incr_report_calls(&namespace, &ctx);
+        self.metrics
+            .incr_authorized_hits(&namespace, &ctx, req.actual_amount);
+
+        Ok(Response::new(CommitResponse {
+            reservation_released: commit_resp.reservation_released,
+        }))
     }
 }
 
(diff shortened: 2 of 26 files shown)


limitador-server/src/envoy_rls/kuadrant_service.rs
```diff
@@ -183,6 +211,151 @@ impl RateLimitService for KuadrantService {
...
+        // Both the fraction-of-limit clamp on `amount` and the ceiling on `ttl` happen inside
+        // `RateLimiter::reserve` itself (`RateLimiterBuilder::reservation_limits`, set once
+        // when the limiter is constructed) - the fraction clamp needs each matching counter's
+        // own max_value, already resolved there, so there's no point re-resolving limits here
+        // just to clamp again; `ttl` is simply passed through unclamped (or, if unset, as a
+        // large sentinel the library's own ceiling will cut down to size).
+        let ttl = req
+            .ttl
+            .and_then(|d| Duration::try_from(d).ok())
+            .unwrap_or(UNSET_RESERVATION_TTL);
```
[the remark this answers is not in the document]
eguzki: the clamp is designed to let the call still succeed

limitador/src/lib.rs
```diff
@@ -464,6 +494,96 @@ impl RateLimiter {
...
+
+        counters
+            .iter()
+            .try_for_each(|counter| self.storage.update_counter(counter, actual_amount))?;
+
+        let reservation_released = if counters.is_empty() {
+            false
+        } else {
+            self.storage
+                .release_reservation(&counters, reservation_id)?
```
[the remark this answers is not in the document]
eguzki: This is a real gap and it isn't a small fix. This should be tracked as a follow-up given the scope.

limitador/src/storage/keys.rs
```diff
@@ -39,6 +39,18 @@ pub fn key_for_counter(counter: &Counter) -> Vec<u8> {
...
+// share a key with the counter's own value (a plain string incremented via `INCRBY`) —
+// Redis only allows one data type per key. Deriving this key by suffixing
+// `key_for_counter`'s bytes, rather than serializing the counter's identity again from
+// scratch, avoids a second encode and trivially inherits the `{namespace}` hash tag the
+// counter key already carries, keeping both keys on the same Redis Cluster shard.
+pub fn key_for_reservations(counter: &Counter) -> Vec<u8> {
+    let mut key = key_for_counter(counter);
+    key.extend_from_slice(b",reservations");
+    key
+}
```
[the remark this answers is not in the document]
eguzki: This issue predates this PR. Will be addressed in separate PR

limitador/src/storage/local_reservations.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl LocalReservationRegistry {
+    pub(crate) fn new(max_size: u64) -> Self {
+        Self {
+            entries: CacheBuilder::new(max_size)
+                .expire_after(ReservationExpiry)
+                .build(),
+            admission_lock: Mutex::new(()),
+        }
+    }
```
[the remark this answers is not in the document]
eguzki: The trade-off is reasonable. Live entries will not be retained until expiry

limitador/src/storage/mod.rs
```diff
@@ -289,6 +336,37 @@ pub trait CounterStorage: Sync + Send {
...
+    /// `counters`, if a live entry is found. Returns whether anything was released.
+    fn release_reservation(
+        &self,
+        _counters: &[Counter],
+        _reservation_id: &ReservationId,
+    ) -> Result<bool, StorageErr> {
+        Err(StorageErr::unsupported(
+            "reservations are not supported by this storage backend",
+        ))
+    }
```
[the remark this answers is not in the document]
eguzki: This is deliberately left unimplemented. Distributed reservation is unsupported today.

limitador/src/storage/local_reservations.rs
```diff
@@ -0,0 +1,200 @@
...
+                };
+                if now_empty {
+                    self.entries.invalidate(counter);
+                } else {
+                    self.entries.insert(counter.clone(), entry);
+                }
+            }
+        }
+        released
+    }
```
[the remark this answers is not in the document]
eguzki: Fixed

limitador/src/lib.rs
```diff
@@ -464,6 +494,96 @@ impl RateLimiter {
...
+                limit_name: None,
+            });
+        }
+
+        let amount = counters
+            .iter()
+            .map(|c| (c.max_value() as f64 * self.reservation_limits.max_fraction) as u64)
+            .min()
+            .map_or(amount, |max| amount.min(max));
+        let ttl = ttl.min(self.reservation_limits.max_ttl);
```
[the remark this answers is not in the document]
eguzki: Fixed

silvi-t (commented):
Verified the PR by running through the verification steps using the sandbox environment. All steps passed.

**Minor note**: The `admission_lock` field doc in `local_reservations.rs` (L46-50) says `release()` doesn't need the lock, but `release()` does acquire it (L185) with a comment explaining why. The two comments contradict each other.

Not an expert in Rust or the Limitador codebase — LGTM based on successful verification.

eguzki:
> Minor note: The admission_lock field doc in local_reservations.rs (L46-50) says release() doesn't need the lock, but release() does acquire it (L185) with a comment explaining why. The two comments contradict each other.

Good catch. The admission_lock was added to the release() call lately to prevent some race issues. Fixing it

limitador/src/storage/redis/scripts.rs
```diff
@@ -56,3 +56,100 @@ pub const VALUES_AND_TTLS: &str = "
...
+
+    for i = 1, n do
+        local counter_key = KEYS[3 * i - 2]
+        local reservation_key = KEYS[3 * i - 1]
+        local limit_key = KEYS[3 * i]
+        local window_secs = tonumber(ARGV[2 * i - 1])
+        local max_value = tonumber(ARGV[2 * i])
+
+        if redis.call('exists', counter_key) == 0 then
+            redis.call('set', counter_key, 0, 'EX', window_secs)
```
adam-cattermole, line 111: As far as I understand this reserve is pre-creating the key at time 0 for the window. When the `SCRIPT_UPDATE_COUNTER` sets the value, if it's a fresh key and it does `INCRBY` it checks that the val is the delta it just incremented by and updates the expiry to the TTL, effectively extending the window. This differs from the in-memory variant where 'touch' starts a window if one hasn't started yet. In this redis/ async redis case the first update to the counters after a reserve was called and the counter is still 0 will result in the TTL being refreshed/re-anchored to this time

limitador/src/lib.rs
```diff
@@ -669,6 +804,101 @@ impl AsyncRateLimiter {
...
+                counters,
+                limit_name: None,
+            });
+        }
+
+        let amount = counters
+            .iter()
+            .map(|c| (c.max_value() as f64 * self.reservation_limits.max_fraction) as u64)
+            .min()
+            .map_or(amount, |max| amount.min(max));
```
adam-cattermole, line 850: Right now this is throwing me a little from the API design / usage point of view. If I understand correctly; if the limit is 10, the reserve call of 1000 will silently get clamped (with default `max_fraction 1.0` to 10) - the reservation succeeds with Ok, but the caller has no idea how much was reserved. The commit is also unconditional so when it does consume the 1000 they asked for it get's stored

limitador-server/proto/kuadrantrls.proto
```diff
@@ -3,11 +3,48 @@ syntax = "proto3";
...
+  repeated envoy.extensions.common.ratelimit.v3.RateLimitDescriptor descriptors = 2;
+  uint64 amount = 3;
+  google.protobuf.Duration ttl = 4;
+}
+
+message ReserveResponse {
+  envoy.service.ratelimit.v3.RateLimitResponse.Code code = 1;
+  // Only set when code is OK. Must be passed back to Commit to resolve this reservation.
+  string reservation_id = 2;
+}
```
adam-cattermole, line 45: I'm wondering, (semi related to https://github.com/Kuadrant/limitador/pull/515#discussion_r3959861791), whether we would want this to be more expressive. I was thinking given that the `reservation_id` is always present in the case that we're OK, could we leverage proto `oneof` to be more descriptive. i.e. if it's the oneof OK variant it'd hold the `reservation_id` and perhaps `granted_amount` actually reserved (in the case of clamp). The `OverLimit` variant could then hold `response_headers_to_add` to support RLS headers down the line - thinking out loud here, curious what you think

limitador/src/lib.rs
adam-cattermole: Thinking more this should deny if over `max_fraction` instead of silently clamping right?

limitador/src/lib.rs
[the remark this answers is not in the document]
adam-cattermole: @eguzki yeah this seems somewhat severe - there's the gap between the commit and release where both the reservation and the counter update is present, but not only that the `?` on
```
counters
    .iter()
    .try_for_each(|counter| self.storage.update_counter(counter, actual_amount))?;
```

fails fast leaving any errored updates to hold the reservation until ttl

limitador-server/src/envoy_rls/kuadrant_service.rs
```diff
@@ -1,23 +1,51 @@
...
 use crate::prometheus_metrics::PrometheusMetrics;
 use crate::Limiter;
 use limitador::limit::Context;
+use limitador::reservation::ReservationId;
+
+// Used only when the caller doesn't set `ttl` at all. Deliberately not "the real default" -
+// the library's own `RateLimiterBuilder::reservation_limits` ceiling (set once, at
+// construction time) always clamps this down to whatever's actually configured, so this only
+// needs to be "large enough to never be the binding constraint," not accurate.
+const UNSET_RESERVATION_TTL: Duration = Duration::from_secs(u32::MAX as u64);
```
adam-cattermole: nit: instead of a const 'sentinel' duration value that'll never be used in practice, what if the `reserve` just took an `Option<Duration>` and you either pass the duration if provided or not

limitador-server/src/main.rs
```diff
@@ -92,20 +93,30 @@ impl From<LimitadorError> for LimitadorServerError {
...
     pub async fn new(config: Configuration) -> Result<Self, LimitadorServerError> {
+        let reservation_limits = ReservationLimits {
+            max_fraction: config.max_reservation_fraction,
+            max_ttl: config.max_reservation_ttl,
+        };
         let rate_limiter = match config.storage {
-            StorageConfiguration::Redis(cfg) => Self::redis_limiter(cfg).await,
-            StorageConfiguration::InMemory(cfg) => Self::in_memory_limiter(cfg),
+            StorageConfiguration::Redis(cfg) => Self::redis_limiter(cfg, reservation_limits).await,
+            StorageConfiguration::InMemory(cfg) => Self::in_memory_limiter(cfg, reservation_limits),
```
adam-cattermole, line 102: nit: not the end of the world, but I wonder if we could've implemented a `From` or some struct for the config so it was all passed in one place rather than the additional param - doesn't really matter though

limitador/src/lib.rs
```diff
@@ -464,6 +494,110 @@ impl RateLimiter {
...
+    /// means a late or already-expired reservation degrades gracefully to plain usage
+    /// reporting rather than failing.
+    pub fn commit_reservation(
+        &self,
+        namespace: &Namespace,
+        ctx: &Context,
+        reservation_id: &ReservationId,
+        actual_amount: u64,
+    ) -> LimitadorResult<CommitResult> {
+        let counters = self.counters_that_apply(namespace, ctx)?;
```
adam-cattermole, line 599: I realise this isn't going to happen in our case, but wonder if it's worth doc'ing that a commit that resolves fewer limits than the original reserve leaves those stranding until TTL

limitador/src/storage/redis/scripts.rs
eguzki: Good catch. Fixed

eguzki: My fix relies on having Redis >= 7.0, It was released in 2022. According to the supported configurations for the product, https://access.redhat.com/articles/7092611, Redis Entrerprise and Cloud  is supported, which in turn they support Redis 6.2, 7,2, 7.4, 8.X. 

Redis 6.2 is near EOL (April 2027). So I am taking my chances here. wdyt?

adam-cattermole: I'm glad you looked into this - I actually recall this from before, maybe here where we tried to keep support for the older versions https://github.com/Kuadrant/limitador/pull/281#discussion_r1539016256.

Honestly I think I would try to avoid breaking backward compatibility unless we are happy issuing a warning that upgrading could break things, I suspect a lot of users don't leverage redis, but concerned that those that do might be on older versions

limitador/src/lib.rs
eguzki: IMO The reservation system's job is to stop multiple concurrent in-flight requests from collectively overcommitting the limit. Not to bound what a single request's real usage turns out to be later.

In your example (max_value=10, current 0, reserve(1000), max_fraction=1): the clamp holds 10. The entire limit and admits it. That's the correct outcome: outstanding is now 10, so any other concurrent reserve() is denied (0+10+anything>10) until this one commits or expires. The one thing this mechanism promises, no other in-flight request gets admitted while this hold is live, is satisfied.

commit's later actual_amount=1000 pushing the counter past max_value isn't a gap in the reservation logic; it's the same thing that happens with plain Report/check_and_update whenever a single request's real usage is much larger than expected. Reserve/commit doesn't (and can't) prevent that. It only prevents other requests from compounding it while this one is outstanding.

I would not turn max_fraction from a clamp into a per-call admission cap, unless I miss something or get this entirely wrong.

limitador-server/proto/kuadrantrls.proto
eguzki: Good idea on oneof in general, but I'd split this into two separate questions:

- granted_amount in the OK variant: commit's actual_amount comes from the caller's own real usage, independent of what was granted. What is the value of surfacing oustanding amount per call? 
- oneof + response_headers_to_add for OverLimit: This makes sense to me. Do you want me to include it in this PR?

limitador/src/storage/redis/scripts.rs
eguzki: Asked internally and I'll get back to this once I have an answer.

eguzki: > I'm glad you looked into this - I actually recall this from before, maybe here where we tried to keep support for the older versions https://github.com/Kuadrant/limitador/pull/281#discussion_r1539016256.

Yeah, my comment  `Actually, I think it is not needed, as the key has just been created, so expiration has not been set` is no longer true. The reservation call can create a counter with 0 value and set the expiration.

limitador/src/lib.rs
adam-cattermole: Sure I understand it's for multiple requests rather than single - but this seems like a strange API/interaction design choice - obviously the commit can overshoot, that we know and is expected. From the point of view of the caller though, they requested to reserve x number of tokens, were subsequently told that was okay (without ever knowing the size of your reservation), and consume an arbitrary amount afterwards through commit, regardless of how far past the limit their original reservation request was. To me this seems like you're 'lying' to the caller.

limitador-server/proto/kuadrantrls.proto
adam-cattermole: > granted_amount in the OK variant: commit's actual_amount comes from the caller's own real usage, independent of what was granted. What is the value of surfacing oustanding amount per call?

The user has some indication for observability or telemetry purposes how much past the limit their request was by seeing how much they asked for vs how much was granted. Otherwise in the okay variant there's no way you could possibly know how much the reservation is for - from the point of view of the libraries API, it seems like this could be important information to surface to the caller? this is only really an issue because of the clamping - the API 'lies' to the caller saying the reservation was accepted regardless of how much past budget the original request was

limitador/src/storage/redis/scripts.rs
eguzki: Product-wise we are safe here. Since RHCL v1.2 (kuadrant 1.3), tested and supported versions are Redis >= 8.2

I would say it is safe to rely on the expire NX feature introduced in Redis 7.0

limitador-server/src/envoy_rls/kuadrant_service.rs
eguzki: Always felt wrong

Fixed as you suggested.

limitador-server/src/main.rs
eguzki: Fair nit. Kept reservation_limits as a separate param on purpose.  it's a limiter-wide policy, not per-storage backend. I try to avoid duplicating aroung for each  existing *StorageConfiguration struct

Happy to leave as-is unless you feel strongly.

limitador/src/lib.rs
eguzki: Good catch, and yes, this is an accepted tradeoff from how the data structure was designed, not an oversight.

Two options were considered for tracking outstanding reservations:
1. **(Implemented)** `commit` takes descriptors + `reservation_id`; both cleanup (`release`) and accounting (`update_counter`) operate on the counters resolved from those descriptors *at commit time*.
2. **(Rejected)** `commit` takes only `reservation_id`; a second map keyed by `reservation_id -> counters` would let both cleanup and accounting target the exact counters `reserve` originally held, independent of how the descriptors resolve later.

I went with (1) because accounting should reflect the currently-configured limits at commit time, not whatever was live when the reservation was made.  Option (2) also means keeping a second index in sync with the first, for a benefit (guaranteed release even under config drift) that only matters for a rare edge case.

The consequence, as you noted: if the resolved counter set changes between `reserve` and `commit` (e.g. limits reconfigured in between), a reservation held on a counter no longer resolved isn't explicitly released. it just sits until its own ttl expires. I'll add a note to `commit_reservation`'s doc comment spelling this out.

limitador/src/storage/redis/scripts.rs
```diff
@@ -11,10 +11,14 @@
...
 // ARGV[2]: delta
+//
+// `expire ... NX` (Redis >= 7.0) skips (re)setting the TTL if one is already running -
+// needed because `SCRIPT_RESERVE` can pre-create a counter key at 0 with a live TTL, and
+// `c == delta` alone can't tell that apart from a genuinely brand new key.
 pub const SCRIPT_UPDATE_COUNTER: &str = "
     local c = redis.call('incrby', KEYS[1], ARGV[2])
     if c == tonumber(ARGV[2]) then
-      redis.call('expire', KEYS[1], ARGV[1])
+      redis.call('expire', KEYS[1], ARGV[1], 'NX')
```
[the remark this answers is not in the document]
eguzki: this is not not simple. First question would be where. In the lib for all consumers or in limitador-server? Second concern would be redis compatible services might report nonstandard version string from the INFO call. 

This is nice to have, but leaving it for latter as follow up work if really needed.

limitador/src/storage/redis/scripts.rs
adam-cattermole: Yep, I'd say lets go for it and document things

limitador/src/storage/redis/redis_sync.rs
```diff
@@ -244,7 +321,71 @@ impl From<::r2d2::Error> for StorageErr {
...
+    // Regression test for the bug fixed by adding `NX` to `SCRIPT_UPDATE_COUNTER`'s
+    // `expire` call: `Reserve` pre-creates a counter's key at value 0 to start its window,
+    // so the first real `update_counter` after it (here, via `Commit`) must not treat that
+    // as "a fresh key" and re-anchor the TTL - the window `Reserve` already started must be
+    // preserved.
+    #[test]
+    #[serial]
+    fn reserve_then_update_counter_does_not_extend_the_ttl() {
+        let storage = RedisStorage::default();
+        storage.clear().unwrap();
```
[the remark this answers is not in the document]
eguzki: This matches the existing convention for every Redis-backed test in this crate

limitador/src/lib.rs
adam-cattermole: Yeah sure, that all makes sense to me, thanks for describing your thoughts - I agree the chosen approach (1) is preferable - doc'ing sounds good

limitador-server/proto/kuadrantrls.proto
eguzki: Would be a good trade-off for you to keep the clamping approach (instead of admission cap) and return reserved amount in the ReserveResponse ?

limitador/src/lib.rs
eguzki: ~Resolving this discussion as it is an (almost) duplicated one happening in this review~

limitador-server/proto/kuadrantrls.proto
adam-cattermole: I think that one or the other approaches would make sense - I'm very undecided as to what _correct_ behaviour would be, a part of me still thinks the deny case makes sense. Even in the concurrent view of the world, as you approach your limit you can still blow your limit by n-1 the reservation amount. On the converse though you also are unlikely to fully consume your whole limit if you reservation is denied beforehand. I guess right now with static values it doesn't make sense, but when the token consumption could be provided as a 'hint' in the request, a request that hints higher max tokens that would make your reservation overshoot could be denied, whereas a smaller request which hints lower usage would be allowed, potentially reducing the overshoot.

Wondering if we could get some other opinions here, maybe @maleck13 or @alexsnaps? (more context in https://github.com/Kuadrant/limitador/pull/515#discussion_r3959861791)

eguzki: I'm considering an alternative: admission checks the raw requested amount (`value + outstanding + requested > max_value → deny`), and if admitted, the amount actually held gets clamped by _max_fraction_ and returned in the response (`reserved_amount`).

For your example (max_value=10, current 0, max_fraction=1, reserve(1000)): admission checks the raw 1000 against the 10 available → 1000 > 10 → the reservation is denied outright, rather than silently admitted holding 10. The caller finds out immediately that the estimate can't be accommodated, instead of getting an Ok that only turns out to represent a fraction of what was asked.

Current implemented behavior would have claimed the full 10 and blocked everyone else until commit/expiry.

eguzki: To be clear: I'm not proposing denial because clamping happens. Denial happens purely because there's no room for the requested amount (`value + outstanding + requested > max_value`). Clamping is a separate step that only ever applies to what gets held once a request is already admitted. It's a self-defense mechanism so that one admitted request can't claim the entire limit for itself, independent of whether more room was technically available.

adam-cattermole: > I'm considering an alternative: admission checks the raw requested amount (value + outstanding + requested > max_value → deny), and if admitted, the amount actually held gets clamped by max_fraction and returned in the response (reserved_amount).

Sure this sounds like a pretty good middle ground to me, I think with the clamp potentially lowering the amount reserved there's still use in the `granted_amount` or similar returning the amount actually reserved to the caller for telemetry / observability

eguzki: going for it then

eguzki:
- Before: the requested amount was clamped by max_fraction before the admission check, so the check itself only ever saw the (possibly much smaller) clamped value. A request far exceeding a counter's max_value would still get silently admitted, holding whatever was left. The caller had no way to tell it wasn't given what it asked for.
- Now: admission checks the raw, requested amount, all or nothing, consistent with how check_and_update/is_rate_limited already work (they never silently apply a smaller delta than requested). `max_fraction` no longer affects whether a reservation is admitted; it only caps how much of an admitted reservation is actually held (`ReserveResult::amount` / `ReserveResponse.reserved_amount`), as a self-defense limit so one reservation can't claim an entire counter.

Threaded the split through the storage layer: CounterStorage/AsyncCounterStorage::reserve now takes check_amount and hold_amount separately, , the Redis Lua script and the local in-memory reservation registry both check with the former and hold with the latter.

Added test coverage across all reservation-capable backends for: over-max_value denial, normal admission, zero-amount checks.

adam-cattermole (commented):
Cool, I think this is looking pretty good now, only outstanding thoughts/questions I have:
- Do we want to go the `oneof` route on the proto for the `ReserveResponse` fields now to save changes in multiple places down the line? adding the headers field and support for it in the `OVERLIMIT` branch would just be additive rather than breaking
- Something I didn't mention in the previous review and I don't necessarily expect to be changed but should potentially be documented somewhere - the reservation layer isn't taken into account by regular rate limiting - so a limit who's remaining quota is reserved by one or more reservations can be consumed as normal until there's no remaining quota before all of the pending reservations then commit. In this scenario you can overshoot by the full limit value if everything is reserved
- https://github.com/Kuadrant/limitador/pull/515#discussion_r3978954811

limitador-server/proto/kuadrantrls.proto
adam-cattermole: Thanks for the changes. I still feel a little uneasy about the `max_fraction` clamping - we'll have to doc this very well I think to avoid problems. It succeeds in it's goal of preventing a single request monopolising the counter:

> It's a self-defense mechanism so that one admitted request can't claim the entire limit for itself, independent of whether more room was technically available.

But it does this by letting through many reservations that all are under-accounted for and will overshoot. The lower the `max_fraction` value, the more requests can be admitted that will overshoot. Not asking for changes, just calling out potential problems in the design

Example with limit 10, (assuming commit consumes amount that was asked for, as we can't control that):
```
max_fraction: 1.0,  hold_cap: 10, reservations: 1  consumed: 10 (1.0x max_value)
max_fraction: 0.5,  hold_cap: 5,  reservations: 2  consumed: 15 (1.5x max_value)
max_fraction: 0.25, hold_cap: 2,  reservations: 5  consumed: 30 (3.0x max_value)
max_fraction: 0.1,  hold_cap: 1,  reservations: 10 consumed: 55 (5.5x max_value) 
```

limitador/src/lib.rs
```diff
@@ -810,4 +1110,281 @@ mod test {
...
+        rl.add_limit(limit);
+
+        let ns = namespace.into();
+        let ctx = Context::default();
+
+        let result = rl
+            .reserve(&ns, &ctx, 1, Some(Duration::from_secs(30)), false)
+            .unwrap();
+        assert!(!result.limited);
+        assert!(result.reservation_id.is_some());
```
adam-cattermole: One point of confusion here - it says in the comment that the clamp to 0 doesn't create a reservation but in this case it still passes back `Some(reservation_id)` instead of `None`, even though this is never stored by the backend. This reservation is also unreleasable by commit and will return `reservation_released: false`

eguzki:
> But it does this by letting through many reservations that all are under-accounted for and will overshoot. The lower the max_fraction value, the more requests can be admitted that will overshoot. Not asking for changes, just calling out potential problems in the design

I was thinking on adding this code doc:
```rust
    /// Caps the amount actually held by an admitted `reserve()` call to
    /// `counter.max_value() * max_fraction`, for every matching counter, the most
    /// restrictive one wins. This never affects admission itself (that's always decided on
    /// the raw requested amount); it only limits how much of a counter a single reservation
    /// can claim, so it can't starve concurrent reservations. Defaults to `1.0`, i.e. no
    /// clamp beyond each counter's own `max_value`.
    ///
    /// This is a trade-off knob, not a free safety win: a lower `max_fraction` admits more
    /// *concurrent* reservations (better for fairness), but each holds a smaller, less
    /// accurate share of the real amount it may end up committing - so the more of them are
    /// in flight at once, the more the counter's real value can overshoot `max_value` once
    /// they all commit. A higher `max_fraction` accepts more monopolization risk (one
    /// reservation can claim more of the counter, or all of it at `1.0`) in exchange for a
    /// tighter bound on aggregate overshoot.
    pub max_fraction: f64,
```

eguzki:
> Do we want to go the oneof route on the proto for the ReserveResponse fields now to save changes in multiple places down the line? adding the headers field and support for it in the OVERLIMIT branch would just be additive rather than breaking

I think that `oneof` isn't required to keep this extensible. protobuf additive fields (a new field number)  are non-breaking. So adding `response_headers_to_add` to `ReserveResponse` later  would be just fine. 

Additionally, RLS response headers aren't OverLimit only, are they?

adam-cattermole:
> I was thinking on adding this code doc:
> --snip--

Sure this seems like a good idea and we could doc it in the operator usage too / wherever we make this user-facing. I realise `max_fraction` is defaulted to `1.0` in the lib but `0.5` in the server, is that intentional?

eguzki:
> Something I didn't mention in the previous review and I don't necessarily expect to be changed but should potentially be documented somewhere - the reservation layer isn't taken into account by regular rate limiting - so a limit who's remaining quota is reserved by one or more reservations can be consumed as normal until there's no remaining quota before all of the pending reservations then commit. In this scenario you can overshoot by the full limit value if everything is reserved

The RFC 0021 already calls this out explicitly (under "Future possibilities"):

```
"a single Limitador counter must never be admitted against under both modes... 
CheckReport traffic would race past the limit exactly as in today's bug, while consuming headroom that Reservation traffic honestly held back for."
```

Worth noting: Limitador itself doesn't enforce this, the RFC relies on the caller (Kuadrant operator's per-policy counter-identity scheme) to guarantee a given counter is only ever used in one mode.

adam-cattermole:
> I think that oneof isn't required to keep this extensible. protobuf additive fields (a new field number) are non-breaking. So adding response_headers_to_add to ReserveResponse later would be just fine.

Sorry, to be clearer, field addition is additive - adding a new `response_headers_to_add` to the current flat response structure is non-breaking. But adopting `oneof` later is breaking and requires moving of current fields into the `oneof` variant that's introduced, whereas it's cheap now - that's why I'm asking if we'd like to go for it up front. Alternatively we could mark `reservation_id` and `reserved_amount` `optional` for now - at the end of the day the only two components that need to be synced are deployed together with the `kuadrant-filter` and the code here, so we could land the breaking change at the same time.

> Additionally, RLS response headers aren't OverLimit only, are they?

Yep you're right, these would be present across all variants unlike the `reservation_id` and `reserved_amount` which only make sense in the Ok variant

limitador/src/lib.rs
eguzki: Good catch. That made me think that on such cases, the commit phase would not have reservation_id. So I have made the reservation_id optional (at both protobuf and rust code)

eguzki:
> Sorry, to be clearer, field addition is additive - adding a new response_headers_to_add to the current flat response structure is non-breaking. But adopting oneof later is breaking and requires moving of current fields into the oneof variant that's introduced, whereas it's cheap now - that's why I'm asking if we'd like to go for it up front. Alternatively we could mark reservation_id and reserved_amount optional for now - at the end of the day the only two components that need to be synced are deployed together with the kuadrant-filter and the code here, so we could land the breaking change at the same time.

Honestly I do not see any hard requirement on having `oneof` types. The wire format and behavior are fully expressible with the current flat + optional fields; nothing is functionally blocked. Even taking into account that headers are landing soon. Checked Envoy's own envoy.service.ratelimit.v3.RateLimitResponse (what CheckRateLimit/Report return) for precedent. It's entirely flat, no oneof.

Anything I might be missing?

adam-cattermole (approved):
> Honestly I do not see any hard requirement on having oneof types. The wire format and behavior are fully expressible with the current flat + optional fields; nothing is functionally blocked. Even taking into account that headers are landing soon. Checked Envoy's own envoy.service.ratelimit.v3.RateLimitResponse (what CheckRateLimit/Report return) for precedent. It's entirely flat, no oneof.

> Anything I might be missing?

Nothing is functionally blocked, agreed, I'm not arguing the wire format can't express it. My point is about what the contract makes representable. With flat and optional, `OverLimit` with `reservation_id: Some(..)` is a well-formed response and it puts the onus on both sides to uphold that `reservation_id` is only meaningful when code is `OK` by inferring, or reading proto comments. With `oneof`, the variant expresses the type, so when code is Ok, you match on the type and extract the value from an Ok variant - the result is the same either way, just one is handled at the type-system level and the other by comments.

On the original `RateLimitResponse`, none of the fields are conditional on the outcome, they're all meaningful regardless of the decision that was made. The closer parallel is envoy's ext_authz [`CheckResponse`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#service-auth-v3-checkresponse) - see `ok_response` and `denied_response` for an example where presence of fields is dependent on the variant returned.

Anyways, I think this makes sense in our case and I like to get APIs right, but I'm not going to block on it and it's changeable later if we decide it's worth tightening the API. Lgtm

eguzki:
Thanks so much for the thorough review on this one, and for being so quick to respond throughout

